PR #7, opened 2026-06-29
fix qaic_utils import

quic-agokhale:
This change fixes the import of qaic_utils from vllm_qaic. The erroneous implementation was trying to import qaic_utils from the vllm package, which would work in the legacy qaic-specific vllm fork but not with the vllm-qaic plugin which relies on vllm cloned from the public repository


ankitqti (approved):
Looks good

PR #9, opened 2026-07-01
Fix links to supported models for AoT and PyTorch modes

quic-agokhale:
This change makes the links on README.md to supported models, point to the vllm-qaic documentation


ankitqti (approved):
Looks good

PR #10, opened 2026-07-02
Enable ngram/suffix/draft-model SpD on v0.23.0

eplatero97:
## Objective
Enable SpD on v0.23.0

## Changes
vLLM v0.23.0 introduced a validation in _set_max_num_scheduled_tokens() that ensures max_num_batched_tokens >= max_num_scheduled_tokens + delta. The EngineCore subprocess re-invokes vllm_config.__post_init__() after the ZMQ handshake, which re-runs the validation.

In __post_init__ ordering, _set_max_num_scheduled_tokens() runs BEFORE the QAIC platform's check_and_update_config(). This means max_num_scheduled_tokens is computed from the original max_num_batched_tokens, then QAIC overrides max_num_batched_tokens to max_num_seqs * long_prefill_token_threshold (a smaller value). When the subprocess re-runs __post_init__, the stale max_num_scheduled_tokens exceeds the overridden max_num_batched_tokens, triggering a ValueError.

Fix: reset scheduler_config.max_num_scheduled_tokens = None immediately after the QAIC max_num_batched_tokens override. None is the field's declared default; _set_max_num_scheduled_tokens() auto-recalculates correctly from the overridden budget on the next __post_init__ invocation.

Also replaces create_vllm_config_for_draft_model (removed in v0.23.0) with vllm.config.utils.replace, and fixes stale tuple-unpacking in _calc_spec_decode_metadata._get_cumsum_and_arange call sites.

## Validation
Successfully an `qaic.py`, `qaic_ngram_suffix.py`, and `qaic_spd.py` files from `examples/`. I also ran all previous unit tests from release `v0.15.0`

### DIFF
--- a/vllm_qaic/platform_base.py
+++ b/vllm_qaic/platform_base.py
@@ -331,6 +331,11 @@ class QaicPlatform(Platform):
                     scheduler_config.max_num_seqs
                     * scheduler_config.long_prefill_token_threshold
                 )
+            # Reset max_num_scheduled_tokens so that
+            # _set_max_num_scheduled_tokens() recalculates it from the
+            # (now-overridden) max_num_batched_tokens when __post_init__
+            # re-runs in the EngineCore subprocess (core.py:1038).
+            scheduler_config.max_num_scheduled_tokens = None
 
         if cls.is_aot:
             # Intel OpenMP tuning — only activate when user has already preloaded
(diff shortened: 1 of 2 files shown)


ankitqti:
Nice work Erick !!

ankitqti (approved):

PR #12, opened 2026-07-06
Improve install UX: config banner and disk-space pre-flight for triton-cpu installation

eplatero97:
## Summary

Installing triton-cpu for AOT SpD hits a silent build failure when default path `$HOME` does not have enough storage to install triton-cpu backend.

This PR makes two improvements to prevent that class of failure from being silent.

---

## 1. Expanded configuration banner (`scripts/install.sh`)

The startup banner previously showed only mode, install source, python version,
and a single triton-cpu enabled/disabled line.

It now shows the full install configuration before anything is downloaded or
built, grouped by mode:

**AOT mode (TRITON_CPU=1):**
```
========================================================
  vllm-qaic installer
  mode     : aot
  source   : source (/path/to/vllm-qaic)
  python   : Python 3.10.12
  env      : /path/to/.venv
  --------------------------------------------------------
  vllm           : 0.23.0       vllm-qaic: 1.22
  torch (AOT)    : 2.7.0+cpu  torchvision: 0.22.0+cpu
  qeff branch    : release/v1.22.0
  target device  : empty
  triton-cpu     : enabled
    src          : /prj/.../scratch/triton-cpu
    commit       : e60f448f8f197073b75d6d3e77347414a5db3ee7
    max jobs     : 4
  --------------------------------------------------------
  Override any variable before running, e.g.:
    TRITON_CPU=1 TRITON_CPU_SRC=/data/triton-cpu ./scripts/install.sh aot
========================================================
```

The triton-cpu `src` path is now visible upfront, so users can see at a glance
whether it resolves to a quota-limited filesystem before committing to a 30-minute
build.

---

### Pre-flight check (before any git clone)

- Walks up `TRITON_CPU_SRC` to the first existing ancestor directory (the clone
  target won't exist yet) and checks available space with `df -Pk`.
- Errors early with an actionable message if < 10 GB available, including the
  exact `TRITON_CPU_SRC` path and how to redirect it.
- Escape hatch: `TRITON_CPU_SKIP_DISK_CHECK=1`.

### Build error handler (wraps `pip install -e`)

If the C++ build fails for any reason, prints a named explanation covering the
`Disk quota exceeded` / `No space left on device` cases, with specific guidance
on redirecting `TRITON_CPU_SRC` to a quota-free filesystem.

---

### Disk-space check — tight filesystem
Ran with `TRITON_CPU_SRC=/boot/triton-cpu` (`/boot` has 882 MB available).
Script exited 1 before any git clone, printing the actionable error message.
No artifacts were written.

### Disk-space check — adequate filesystem
Default `TRITON_CPU_SRC` path resolves to a filesystem with 637 GB available.
Check passes silently.

### NFS quota scenario (reproduces the original failure)
Ran full `TRITON_CPU=1 ./scripts/install.sh aot` with `TRITON_CPU_SRC` on
`$HOME` (637 GB filesystem free, but per-user NFS quota exhausted mid-build).
The `df` pre-flight passed (as expected — `df` cannot see user quota), and the
build failed with `[Errno 122] Disk quota exceeded`. The new error handler
printed a named explanation and redirect instructions instead of a silent exit.

### Full AOT + triton-cpu install — scratch NFS (no quota)
Ran with `TRITON_CPU_SRC` on a scratch NFS mount (3.8 TB, separate server from
`$HOME`, no per-user quota). All four install steps completed successfully:

| Step | Package | Result |
|------|---------|--------|
| 1 | QEfficient (torch 2.7.0+cpu) | ✓ |
| 2 | vLLM deps + vLLM 0.23.0 | ✓ |
| 3 | vllm-qaic-aot | ✓ |
| 4 | triton-cpu (`OK: cpu backend is active`) | ✓ |

Final install into `.venv/` in the repo root also completed successfully with
the same result.

### DIFF
--- a/docs/docs/getting_started/installation/aot_mode.md
+++ b/docs/docs/getting_started/installation/aot_mode.md
@@ -12,13 +12,26 @@ AOT (Ahead-of-Time) mode uses the QEfficient library to compile models into Qual
 ./scripts/install.sh aot
 ```
 
+!!! info "Configuration banner"
+    Before installation starts, `install.sh` prints a full summary of every version and
+    setting it will use (vllm, vllm-qaic, torch, qeff branch, target device, triton-cpu
+    state). Review the banner output and override any variable before re-running.
+
 ??? example "Optional environment overrides"
     ```bash
     # Pin transformers version
     TRANSFORMERS_VERSION_AOT=4.55.3 ./scripts/install.sh aot
 
     # Enable triton-cpu backend (for Speculative Decoding)
+    # triton-cpu is a large C++ build — requires ~10 GB of free disk space at TRITON_CPU_SRC
     TRITON_CPU=1 ./scripts/install.sh aot
+
+    # Redirect the triton-cpu clone+build to a filesystem with more space
+    # (use when $HOME has limited disk space or a user quota)
+    TRITON_CPU=1 TRITON_CPU_SRC=/path/with/more/space/triton-cpu ./scripts/install.sh aot
+
+    # Skip the disk-space pre-flight check entirely
+    TRITON_CPU=1 TRITON_CPU_SKIP_DISK_CHECK=1 ./scripts/install.sh aot
 
     # Install from pre-built wheel
     VLLM_QAIC_INSTALL_SOURCE=wheel \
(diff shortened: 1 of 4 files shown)


ankitqti:
Also add about this in docs/installation.md and same in installation section of the documentation

ankitqti (approved):
Looks good

PR #20, opened 2026-07-21
Rework prefill_seq_len sourcing, logits dtype handling, and RMS-norm …

quic-agokhale:
…kernel ABI

- Require prefill_seq_len from override_qaic_config instead of deriving it from scheduler_config.long_prefill_token_threshold; compute max_num_batched_tokens accordingly in QaicPlatform.
- Resolve logits output dtype from the QPC binding (float16 for mxfp6/mxint8, float32 otherwise) instead of hardcoding float32
- Add QaicUniProcExecutor to size max_concurrent_batches for async scheduling with prefill pipeline stages (kv_producer role).
- Simplify rms_norm_hexagon: pass tensors and scalar args directly to the kernel instead of pre-flattening on host and packing params into a float tensor; fix rms_norm_multi_nsp_bf16 to also write the residual output buffer, which was previously dropped.
- Disable the multimodal processor cache; fix mm_token_type_ids shape for decode, and whisper's decode input_features shape.
- Add split_model_io to vision-language examples and gemma4/qwen3_6_vl demos.
- Misc fixes: kv_caches None check, qaic_config/pretrained_extra_args handling, decode_batch_inputs_by_k sync, envs import, mdp_compiler_dump_path.

### DIFF
--- a/csrc/add_rms_norm/dispatch.cpp
+++ b/csrc/add_rms_norm/dispatch.cpp
@@ -37,16 +37,18 @@ extern "C" uint32_t rms_norm_multi_nsp_bf16(const AicJitEntryPointConfig *cfg,
 
 // rms_norm_dispatch — single entry point called by the wrapper.
 // Pointer layout (same as the individual kernels):
-//   ptrs->pointers[0] : attn_out  [M, N]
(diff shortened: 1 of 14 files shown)


ankitqti (approved):
Looks good !!

PR #23, opened 2026-07-23
Add Rotary Embedding support

quic-agokhale:
This change adds support for Rotary Embedding through:
- registering OOT forward for when ApplyRotaryEmb is instantiated and invoked
- patching the forward_static() of ApplyRotaryEmb for when it is invoked using class-level access (e.g., in vllm/model_executor/layers/rotary_embedding/base.py)


vllm_qaic/ops/rotary_embedding.py
```diff
@@ -0,0 +1,70 @@
...
+            cos = cos.to(torch.float32)
+            sin = sin.to(torch.float32)
+        out = torch.ops.qaic.rotary_embedding(x, cos, sin, is_neox_style)
+
+        return out
+
+    def forward_oot(self, x, cos, sin, is_neox_style=True, enable_fp32_compute=False):
+        return self.forward_static(x, cos, sin, is_neox_style, enable_fp32_compute)
+
+ApplyRotaryEmb.forward_static = staticmethod(QAicApplyRotaryEmb.forward_static)
```
ankitqti: for custom_ops forward_oot api doesn't need to change forward_static

why we need to define forward_static, you can move whole implementation of forward_static to forward_oot

vllm_qaic/ops/rotary_embedding.py
```diff
@@ -0,0 +1,70 @@
...
+    The function computes x -> silu(x[..., :d]) * x[..., d:]
+    where d = x.shape[-1] // 2.
+
+    Shapes:
+        x: (num_tokens, 2 * d) or (batch_size, seq_len, 2 * d)
+        return: (num_tokens, d) or (batch_size, seq_len, d)
+    """
+
+    @staticmethod
+    def forward_static(x, cos, sin, is_neox_style=True, enable_fp32_compute=False) -> torch.Tensor:
```
ankitqti: Only forward_oot override is needed

ankitqti (changes requested):

vllm_qaic/ops/rotary_embedding.py
quic-agokhale: In general, yes, we only need to implement forward_oot. But specifically for ApplyRotaryEmb, I observed that there are two ways in which the layer is invoked by higher-level modules.
First is the classic way which would be dispatched by custom op registry to forward_oot()
Second is seen here https://github.com/vllm-project/vllm/blob/0fc695fc6d1d82e9a5ac6835ac8e4e1c83703665/vllm/model_executor/layers/rotary_embedding/base.py#L180 where ApplyRotaryEmb::forward_static() is invoked directly.

To handle both cases, I have routed forward_oot through an overriden forward_static. the same forward_static is patched to the ApplyRotaryEmb class

ankitqti: Then will need customop registry for RotaryEmbedding where we implement forward_oot

ankitqti:
Rebase this PR

DCO issue (-s option not used while creating git commit)

PR #33, opened 2026-07-28
[CI][v0.23.0] Add initial CI setup for AoT

animeshk08:
## Summary

Adds the initial CI test infrastructure for the AoT inference mode: a full e2e pytest suite, a custom test scheduler that replaces `pytest-xdist` for this plugin's needs, and the shell entry points that wire it into CI. Also includes a small number of production-code fixes surfaced while building/running this suite.

### 1. e2e test suite (`tests/`)
- `tests/conftest.py`: generic `VllmRunner` wrapper (adapted from upstream vLLM's own test conftest) — no QAIC specifics.
- `tests/e2e/conftest.py`: the QAIC-specific layer — device pool management, the `qaic_test_config`/`qaic_aot_mode`/`qaic_disagg_installed` pytest markers, and the `qaic_model`/`make_runner`/`server_runner` fixtures tests build on.
- Coverage added: core generation/accuracy/samplers/benchmark/offline-stability/output-consistency, OpenAI-compatible API server + multi-client, LoRA (`tests/e2e/lora/`), multimodal/dual-QPC (`tests/e2e/multimodal/`), and disaggregated prefill/decode serving (`tests/e2e/disaggregated_serving/`).
- `tests/README.md`: short guide for developers adding tests here.

### 2. Custom CI scheduler (`ci_scripts/`)
`pytest -n <N> --dist=loadscope` doesn't understand QAIC's device-pool constraints or its model-export/QPC-compile caching, so this plugin's tests need their own dispatcher.

- `collect_jobs.py`: runs `pytest --collect-only`, groups results into "jobs" (one per class/module scope), and records each job's device requirement, a compile-cache correlation key, and base CLI args. Tests already marked `skip` by the conftest hooks are dropped before job creation.
- `scheduler.py`: dispatches each job as its own `pytest` subprocess with a disjoint slice of real device IDs from an in-memory pool. Serializes the first job per model (export) and per exact config (compile) so concurrent jobs don't race on the same on-disk QEfficient cache directory — later jobs sharing that key reuse the warm cache instead of recompiling.
- `ci_fasttest_qaic_aot.sh`: discovers/accepts a device pool, runs `collect_jobs.py` then `scheduler.py`, and does a post-run device health check.
- `ci_pipeline_run_plugin.sh`: installs the plugin + test deps for a given inference mode, then sources the matching `ci_fasttest_qaic_<mode>.sh` (currently `aot` only).

### 3. New pytest markers (`pyproject.toml`)
```toml
[tool.pytest.ini_options]
markers = [
    "qaic_test_config(**kwargs): per-test/module QAIC run configuration overrides",
    "qaic_aot_mode: test requires QAIC AOT (ahead-of-time compiled) mode to be installed",
    "qaic_disagg_installed: test requires the qaic_disagg package to be installed",
]
```
4. Small production fixes surfaced by this work
* vllm_qaic/platform.py: added get_punica_wrapper() → PunicaWrapperCPU, needed for the new LoRA tests to exercise vLLM's LoRA-batching path on this platform.
* vllm_qaic/patch/patch_parallel_state.py: the cleanup_dist_env_and_memory monkeypatch patched vllm.v1.engine.core's bound reference but missed vllm.distributed's own re-export — some call sites were still hitting the unpatched original. Both are patched now.
* vllm_qaic/patch/patch_config.py: QaicCacheDType was CacheDType | Literal["mxint8"], which breaks if "mxint8" does not exists in upstream's CacheDType. Now builds a deduplicated union via get_args.
* pyproject.toml: corrected license/authors metadata; added [tool.ruff] line-length = 90 and excluded tests/**; added the markers block; fixed a missing trailing newline.

## Why this design?
Device-pool + compile-cache races were a real, reproduced bug during development — two concurrent jobs needing a fresh compile for marker-equivalent-but-not-identical configs could pick the same cache dir and corrupt each other's output. scheduler.py's per-model/per-config gating exists specifically for this.
qaic_model is deliberately scope="class" — compiling to QPC is expensive, so tests sharing a config should share one class/compile rather than one-per-function.
Markers replace ad-hoc importorskip/platform checks with one declarative, centrally-enforced mechanism.

## For reviewers
* Start with tests/README.md for test-authoring conventions before diving into tests/e2e/conftest.py.
* The four vllm_qaic/pyproject.toml production changes are independently reviewable — happy to split into a separate PR if preferred, but each was surfaced directly by this suite so is included here for traceability.
* No existing test or runtime behavior changes outside LoRA/cleanup-patch/cache-dtype paths touched above.

# Results:

```
[scheduler] 11/11 job(s) passed
[scheduler] total execution time: 745.3s (0:12:25)
```


ci_scripts/ci_fasttest_qaic_aot.sh
```diff
@@ -0,0 +1,103 @@
...
+fi
+
+OUTPUT_DIR="$SCRIPT_DIR/scheduler_output/fasttest"
+mkdir -p "$OUTPUT_DIR"
+JOBS_FILE="$OUTPUT_DIR/jobs.json"
+
+echo "[ci_fasttest_qaic] Collecting jobs from: tests/e2e"
+DEVICE_POOL_SIZE=$(($(tr -cd ',' <<< "$DEVICE_IDS" | wc -c) + 1))
+python3 "$SCRIPT_DIR/collect_jobs.py" tests/e2e --output "$JOBS_FILE" -- \
+    --ignore=tests/e2e/multimodal \
```
animeshk08: Currently, many multi-modal tests were failing hence, ignoring them for now. We should revisit this later, also to check what models to include as part of fast ci.

tests/e2e/dataset/audio.wav
ankitqti, line 1: Is possible to get the audio file from vllm repo or from online url ?

vllm_qaic/patch/patch_config.py
```diff
@@ -21,7 +20,10 @@
 
 logger = init_logger(__name__)
 
-QaicCacheDType = CacheDType | Literal["mxint8"]
+_QAIC_CACHE_DTYPES = ("mxint8",)
```
animeshk08, line 23: vllm_qaic/patch/patch_config.py: QaicCacheDType was CacheDType | Literal["mxint8"], which breaks if "mxint8" does not exists in upstream's CacheDType. Now builds a deduplicated union via get_args.

This is flaky as resolves to either mxint8 or upstream CacheDType based on whether plugin is registered first or upstream execution. Hence had to dedup it to a single union of both 

Error thrown in vllm bench

```
vllm bench <bench_type> [options] latency: error: argument --kv-cache-dtype: invalid choice: 'auto' (choose from mxint8)

```

pyproject.toml
```diff
@@ -114,4 +114,11 @@ override-dependencies = [
 ]
 
 [tool.uv.sources]
-vllm = { git = "https://github.com/vllm-project/vllm", rev = "v0.15.0" }
\ No newline at end of file
+vllm = { git = "https://github.com/vllm-project/vllm", rev = "v0.15.0" }
```
animeshk08: TODO? Didnt touch this as I was unsure what this is used for..

tests/e2e/dataset/audio.wav
animeshk08: Done

animeshk08:
Moving to draft to find alternative for open_orca_gpt4_tokenized_llama3.sampled_1024new_truncated_prompts_to_128.pkl which is ignore by gitignore

ankitqti:
@animeshk08  Lets Rebase this patch

Looking good for merge, we will add separate patches for fixes other stuff, at least basic CI infra will be in after this

animeshk08:
> Animesh Kumar (@animeshk08) Lets Rebase this patch
> 
> Looking good for merge, we will add separate patches for fixes other stuff, at least basic CI infra will be in after this

Done

ankitqti (approved):
Good to merge

PR #37, opened 2026-07-28
Remove engine args to enable dict-based support for eager mode vision…

Neeharika-Gupta:
Summary
Enable dict-based support in the eager mode vision language example.

Changes
Remove dependency on engine args for LLM construction.


quic-agokhale:
LGTM! Similar change would be needed for AoT example script as well

quic-agokhale (approved):

PR #40, opened 2026-07-29
[v0.23.0][Bug] Fix async-scheduling deadlock in model runner

animeshk08:
### Summary
`vllm-qaic`'s AoT (ahead-of-time) model runner deadlocks when `async_scheduling=True`, hanging for ~300s before raising a `RuntimeError`. This PR fixes the deadlock and re-enables async scheduling for the AoT generative case.

### Root cause
QAIC's async `sample_tokens()` returns a cheap output wrapper immediately; the expensive work (`complete_all_inf`'s blocking hardware wait, logits computation, sampling, bookkeeping) only runs later inside `get_output()`. The engine calls `get_output()` on batch A only *after* appending A's `(execute_model, sample_tokens)` pair to its `batch_queue`.

With `max_concurrent_batches=2`, this created a circular wait:
- Batch B's `execute_model()` blocks in `synchronize_input_prep()`, waiting on `model_runner_output_event`.
- That event is only ever set inside batch A's `get_output()`.
- The engine won't call `A.get_output()` until B is appended to the queue.
- B can't be appended until its `execute_model()` returns.

→ deadlock → ~300s timeout → `RuntimeError`.

### Fix
- Replace the `Event`-based wait/clear handshake with an explicit `_pending_output` reference on the model runner.
- Make both async output wrapper classes (`QaicAsyncPoolingModelRunnerOutput`, `QaicAsyncGPUModelRunnerOutput`) cache their computed result (`self._output`), so `get_output()` is idempotent and safe to call more than once.
- `synchronize_input_prep()` now eagerly drains the previous batch's pending output (by calling `get_output()` directly) only if one is still outstanding — instead of blocking on an event with a fixed timeout.
- Whichever caller reaches `get_output()` first — the eager drain or the engine's own `future.result()` — does the real work; the other call is a cached no-op. This preserves the device/host overlap the async design is meant to provide (batch B's dispatch can still be issued before A's hardware wait is forced), while eliminating the deadlock.
- Drop the `QaicPlatform` override that unconditionally disabled `async_scheduling` for AoT models, now that the underlying hang is fixed.

### Testing
* Async scheduling test
```
==============================================================
  Iteration 3 — Async Scheduling KPI Comparison
==============================================================
  Metric                            non-async          async
  --------------------------------------------------------
  Avg TTFT (ms)                        138.8          122.3
  Avg Decode Time (ms)                2311.4         2137.7
  Avg E2E Latency (ms)                2450.3         2260.0
  Avg TPOT (ms/tok)                    13.94          13.28
  Throughput (tok/s)                   208.0          227.0
  Decode Throughput (tok/s)            210.8          229.4
  Avg Output Tokens                    190.0          190.0
  --------------------------------------------------------
  Total Exec Time (s)                  10.98          10.21
==============================================================
```

* Prefill-streaming test
```
=======================================================================
  Prefill Streaming Benchmark — no-stream iter3 vs stream iter3
=======================================================================
  Metric                            no-stream iter3       stream iter3
  ---------------------------------------------------------------------
  TTFT mean                                   81.33              51.97  (ms)
  TTFT median                                 50.29              46.37  (ms)
  TTFT p99                                   213.13             110.00  (ms)
  TTFT (from arrival) mean                   433.10             297.52  (ms)
  TTFT (from arrival) median                 422.64             292.04  (ms)
  TTFT (from arrival) p99                    817.36             563.01  (ms)
  E2E mean                                    81.33              51.97  (ms)
  E2E median                                  50.29              46.37  (ms)
  E2E p99                                    213.13             110.00  (ms)
  Throughput                                  93.48             175.81  (tok/s)
=======================================================================

```

### DIFF
--- a/vllm_qaic/platform_base.py
+++ b/vllm_qaic/platform_base.py
@@ -460,20 +460,6 @@ class QaicPlatform(Platform):
                     vllm_config, model_config, scheduler_config, model_type
                 )
 
-        if cls.is_aot and scheduler_config.async_scheduling:
-            logger.warning(
-                "QAIC currently does not support async scheduling; "
-                "Falling back to non-async scheduling."
-            )
-            scheduler_config.async_scheduling = False
-
-        if cls.is_aot and scheduler_config.async_scheduling:
-            logger.warning(
-                "QAIC currently does not support async scheduling; "
-                "Falling back to non-async scheduling."
-            )
-            scheduler_config.async_scheduling = False
-
     @classmethod
     def is_pin_memory_available(cls) -> bool:
         logger.warning("Pin memory is not supported on Qaic.")
--- a/vllm_qaic/worker/model_runner.py
+++ b/vllm_qaic/worker/model_runner.py
@@ -15,7 +15,6 @@ from contextlib import AbstractContextManager, contextmanager, nullcontext
 from copy import copy, deepcopy
 from dataclasses import dataclass
 from queue import Queue
-from threading import Event
 from typing import TYPE_CHECKING, NamedTuple
 
 import numpy as np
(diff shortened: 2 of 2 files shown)


ankitqti (approved):
Looks good

PR #41, opened 2026-07-29
[v0.23.0][AOT] Align AOT dependencies with mainline vllm v0.23.0

animeshk08:


requirements/vllm_dependency_aot.txt
```diff
@@ -19,7 +19,7 @@ py-cpuinfo
 transformers == 5.5.4 # required for QEfficient with tag 'r1.21.0'
 tokenizers >= 0.21.1  # Required for fast incremental detokenization.
 safetensors >= 0.6.2  # MXFP4/MXFP6 dtype support (F8_E8M0, F4) added in 0.6.0: https://github.com/huggingface/safetensors/pull/611
-protobuf==6.31.0 # Required by LlamaTokenizer, gRPC.
+protobuf==6.31.0 # Required by LlamaTokenizer, gRPC. efficient-transformer pins to 6.31.0; vLLM v.23.0 avoid this version to fix CVE-2026-0994(https://github.com/vllm-project/vllm/pull/34253)
```
ankitqti: Lets have same settings as 

protobuf >= 5.29.6, !=6.30.*, !=6.31.*, !=6.32.*, !=6.33.0.*, !=6.33.1.*, !=6.33.2.*, !=6.33.3.*, !=6.33.4.*

Does the package install if we revert to original vllm 0.23.0 setting ?

We will ask QEFF to change due to https://github.com/advisories/GHSA-7gcm-g887-7qv7,

File Jira for QEFF team to fix this version

ankitqti (changes requested):

requirements/vllm_dependency_aot.txt
animeshk08: Reverted to upstream v0.23.0 version. Create Jira with qeff team to request them to update protobuf version as well.

ankitqti: Does the install.sh aot works, with this flow ? Is installation flow broken till this is not fixed in QEFF?

ankitqti: I think we should add "smg-grpc-servicer >= 0.4.2" in dependency list to give to user an option to enable grpc via vllm serve --grpc

animeshk08: > Does the install.sh aot works, with this flow ? Is installation flow broken till this is not fixed in QEFF?

Yes with install.sh and qeff installation work. Ran vllm e2e also works. Only a compatibility warming is thrown during installation

```
qefficient 1.22.0 requires protobuf==6.31.0, but you have protobuf 7.35.1 which is incompatible.
```

> I think we should add "smg-grpc-servicer >= 0.4.2" in dependency list to give to user an option to enable grpc via vllm serve --grpc

Installing smg-grpc-servicer >= 0.4.2 (note: 0.4.2 is the lowest available version), leads to upgrading torch to 2.13 due to its dependency. I think we can avoid adding it in this release.

ankitqti (approved):
Looks good to me

PR #43, opened 2026-07-29
Add fix for guided decoding

quic-sanising:
On QAIC, the upstream function builds the `xgrammar` index tensor with `pin_memory=True` (no CUDA allocator -> crash) and passes a `torch.Tensor` to the CPU xgrammar kernel, which only accepts a python `Sequence[int]`. This breaks guided/structured decoding.

Replace `apply_grammar_bitmask` with a version that gates the pinned tensor path on `is_pin_memory_available()` and passes a plain list of indices on QAIC.


vllm_qaic/utils/qaic_utils.py
```diff
@@ -161,3 +171,87 @@ def _clean_config(
             else:  # For other compiler args
                 update_cfg[key] = value
     return update_cfg
+
+
+def qaic_apply_grammar_bitmask(
```
ankitqti: WHy we are having same function in both utils and patch file ?

quic-sanising: Removed duplicate implementation of the function. Thanks for pointing.

vllm_qaic/utils/qaic_utils.py
ankitqti, line 5: Add apache license to this file as we have copied code from open-source, see model_runner.py for the format of apache license

vllm_qaic/utils/qaic_utils.py
```diff
@@ -161,3 +171,87 @@ def _clean_config(
             else:  # For other compiler args
                 update_cfg[key] = value
     return update_cfg
+
+
```
ankitqti: Lets move this to patch_structured_output.py

ankitqti (changes requested):

vllm_qaic/utils/qaic_utils.py
quic-sanising: Done

vllm_qaic/utils/qaic_utils.py
quic-sanising: This function is needed for both AoT and PyT. If we move it to patch file, it will cause issues during import. 

Because as soon as `patch_structured_output.py` is imported, it does the two monkeypatches at module level. It patches `apply_grammar_bitmask` and `gpu_model_runner`. 

If we import this function here or directly in `model_runner.py` from the `patch_structured_output.py` file for AoT mode, AoT would end up applying the PyT monkeypatches, which we don't need.

In my opinion, we should keep the function here: it works for AoT, and for PyT, we import it in the patch file and that works too.

quic-sanising:
@ankitqti This issue is already fixed in public vLLM v0.23.0: https://github.com/vllm-project/vllm/blob/releases/v0.23.0/vllm/v1/structured_output/utils.py#L44. Therefore, monkeypatching is not needed. Closing this PR.

vllm_qaic/utils/qaic_utils.py
quic-sanising: Done

PR #44, opened 2026-07-29
Allow user to set min and max pixels in override_qaic_config using mm_processor_kwargs

quic-sanising:


ankitqti (approved):
Looks good to me

PR #45, opened 2026-07-29
Convert `image_grid_thw` from 3-D to 2-D for Qwen-VL models

quic-sanising:
In `qaic_disagg`, `_merge_embeds` adds a leading batch dim to `image_grid_thw` `([N, 3] -> [1, N, 3])` and calls `_get_mm_fields_config` directly, so `prod(-1)` is 2-D and `flat_from_sizes` raises `"size_per_item should be a 1-D tensor"` (caught, but logs a traceback per request).

Add a small mixin class that wraps `_get_mm_fields_config` to squeeze `ndim==3` `image_grid_thw` back to `ndim==2`.


vllm_qaic/patch/__init__.py
```diff
@@ -71,6 +71,18 @@
 #       Because vLLM defaults to fork-based subprocesses, patching here (in the
 #       main process before fork) is sufficient.
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
```
ankitqti: I'm not happy to patch model , it will also impact eager mode, is there another way for this issue, can we do something in qaic_multimodel or on our qaic_custommm_processor for this issue

quic-sanising: Sure. I have removed the monkeypatch. I have moved the fix into `qaic_custom_mm_processor.py` using a small mixin class. It works the same way. I have tested it on Qwen2.5-VL and it works. Please review again. Thanks.

ankitqti (approved):
Looks good!!

PR #50, opened 2026-07-30
Remove stale qbuffer cache

chaitanyasachdeva:
Remove the persistent qbuffer cache and stale set_buffers calls so np_run registers the current buffers for each execution.


quic-sanising (approved):

ankitqti:
@chaitanyasachdeva  and @quic-sanising  Did you ran the slow test locally to confirm nothing breaks? what about Embedding model tests ?

@quic-sanising We need performance comparision with this and without this patch for two models llama-8b and GOT-OSS , can you explain @chaitanyasachdeva  how to run some portion of our Performance tests

ankitqti (approved):
Looks good to me

PR #51, opened 2026-07-31
[v0.23.0][AOT] Add embedding model support

athavale-shivani:
Adding embedding model support for vLLM 0.23.0 for AOT.

### DIFF
--- a/examples/qaic_score.py
+++ b/examples/qaic_score.py
@@ -7,6 +7,7 @@
 # Adapted from vllm/examples/offline_inference/basic/score.py
 
 from vllm import LLM
+from vllm.config import PoolerConfig
 
 # Score (reranking) example for QAIC.
 # llm.score(query, passages) runs a cross-encoder model that produces a
@@ -42,8 +43,9 @@ def main():
         enforce_eager=True,
         max_num_seqs=4,
         max_model_len=512,
+        pooler_config=PoolerConfig(task="classify"),
         additional_config={
-            "device_group": [5],
+            "device_group": [0],
             "override_qaic_config": {"pooling_device": "cpu", "task": "score"},
         },
     )
@@ -57,8 +59,9 @@ def main():
         enforce_eager=True,
         max_num_seqs=4,
         max_model_len=512,
+        pooler_config=PoolerConfig(task="classify"),
         additional_config={
-            "device_group": [6],
+            "device_group": [0],
             "override_qaic_config": {"pooling_device": "qaic", "task": "score"},
         },
     )
--- a/vllm_qaic/model_loader/qaic.py
+++ b/vllm_qaic/model_loader/qaic.py
@@ -30,7 +30,7 @@ from vllm.model_executor.layers.pooler.seqwise import (
     pooler_for_classify,
     pooler_for_embed,
 )
-from vllm.model_executor.layers.pooler.special import DispatchPooler
+from vllm.model_executor.layers.pooler.special import BgeM3Pooler, DispatchPooler
 from vllm.model_executor.layers.pooler.tokwise import (
     AllPool,
     pooler_for_token_classify,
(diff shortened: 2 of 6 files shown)


examples/qaic_score.py
```diff
@@ -42,8 +43,9 @@ def main():
         enforce_eager=True,
         max_num_seqs=4,
         max_model_len=512,
+        pooler_config=PoolerConfig(task="classify"),
```
ankitqti, line 46: Why scoring example is using classify ?

examples/qaic_score.py
```diff
@@ -57,8 +59,9 @@ def main():
...
```
ankitqti, line 62: Why scoring example is using classify ?

vllm_qaic/model_loader/qaic.py
```diff
@@ -93,28 +93,27 @@ def __init__(
...
                     ),
-                    # TODO: "score" removed: not a valid PoolingTask in upstream vllm 0.23.0
-                    # (DispatchPooler validates keys; "score" was added in the Qualcomm fork)
-                    # "score": pooler_for_classify(
-                    #     pooler_config,
-                    #     pooling=pooling,
-                    #     classifier=classifier,
-                    #     act_fn=None,  # sigmoid not applied; raw logits returned
-                    # ),
+                    # new task added in upstream vllm
```
ankitqti, line 112: What is BgeM3Pooler, please provide more details ?

vllm_qaic/worker/model_runner.py
```diff
@@ -719,14 +697,14 @@ def _pool(
...
-        # For QAIC based pooler OR score/classify tasks (CPU or QAIC),
+        # For QAIC based pooler OR classify task (CPU or QAIC),
         # model outputs are already pooled — wrap directly in ModelRunnerOutput.
         num_reqs_qaicpooler = self.input_batch.num_reqs
         assert num_reqs_qaicpooler == len(self.input_batch.pooling_params), (
             "Either all or none of the requests in a batch must be pooling request"
         )
         hidden_states_qaicpooler = hidden_states[:num_reqs_qaicpooler]
-        seq_lens_qaicpooler = self.seq_lens.cpu[:num_reqs_qaicpooler]
+        seq_lens_qaicpooler = self.optimistic_seq_lens_cpu[:num_reqs_qaicpooler]
```
ankitqti, line 707: What is optimistic_seq_lens_cpu ? why we need this here ?

ankitqti (changes requested):

examples/qaic_score.py
athavale-shivani: In vLLM0.23.0, "score" does not exist as a PoolingTask anymore. classify is the upstream vLLM task that enables llm.score() to work.

examples/qaic_score.py
athavale-shivani: In vLLM0.23.0, "score" does not exist as a PoolingTask anymore. classify is the upstream vLLM task that enables llm.score() to work.

vllm_qaic/model_loader/qaic.py
athavale-shivani: "embed&token_classify" is a new task and is defined in vllm/tasks.py in vLLM v0.23.0. BgeM3Pooler is its handler defined upstream in vllm/model_executor/layers/pooler/special.py. It needs to be explicitly registered in QAIC's DispatchPooler so the task is recognized.

vllm_qaic/worker/model_runner.py
athavale-shivani: optimistic_seq_lens_cpu  is needed to get the sequence lengths of each request which is used to post process the outputs from QAIC pooler. Previously, self.seq_lens used to have this data but in v0.23.0 we seem to be not updating self.seq_lens and they are all set to 0 at this point in the code which leads to an error. optimistic_seq_lens_cpu is being updated and contains the sequence lengths instead, hence I am using that.

ankitqti (approved):
Looks good

PR #53, opened 2026-08-05
Paged Attention support for AOT

aditjadh:
Summary

This PR adds paged (block-based) KV cache support to the QAIC AOT execution path, which enables prefix caching on vLLM v1.

Until now, AOT allocated one contiguous KV region per sequence, sized to the full context length. That meant:

Prefix caching was explicitly unsupported and raised NotImplementedError.
KV memory scaled with max_num_seqs × max_model_len.
Shared prompt prefixes were recomputed on every request.

With paged attention, KV cache is carved into fixed-size blocks managed by vLLM's block manager and addressed on-device through a per-request block table. Requests only occupy the blocks they need, and identical prefixes can be reused across requests.

Configuration
Paged Attention can be enabled by setting --enable-prefix-caching=True. When enabled, the model is compiled in paged mode and KV blocks are sized to the prefill chunk length. When disabled, the existing single-block-per-sequence path is used.
Block count is derived automatically from the model's context length and prefill chunk size, and can be tuned with --num-gpu-blocks-override.

Limitations
Currently, KV block size is tied to long_prefill_token_threshold; it is not independently configurable.

### DIFF
--- a/vllm_qaic/model_loader/qaic.py
+++ b/vllm_qaic/model_loader/qaic.py
@@ -147,6 +147,16 @@ class QaicCausalLM(nn.Module, SupportsLoRA):
         self.ctx_len = model_config.max_model_len
         self.decode_bsz = vllm_config.scheduler_config.max_num_seqs
         self.full_batch_size = vllm_config.scheduler_config.max_num_seqs
+        self.num_gpu_blocks_per_batch = self.ctx_len // self.prefill_seq_len
+        if vllm_config.cache_config.enable_prefix_caching:
+            self.num_gpu_blocks = (
+                vllm_config.cache_config.num_gpu_blocks_override
+                if vllm_config.cache_config.num_gpu_blocks_override
+                else self.ctx_len // self.prefill_seq_len
+            )
+        else:
+            self.num_gpu_blocks = self.decode_bsz
+        self.paged_attention = bool(vllm_config.cache_config.enable_prefix_caching)
         self.prefill_bsz = 1
         self.lora_mode = bool(vllm_config.lora_config)
         self.last_decode = False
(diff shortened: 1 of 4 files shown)


vllm_qaic/platform_base.py
quic-vargupt, line 435: Can you please also check this assertion? In PD disaggregation, this assertion is getting triggered

vllm_qaic/model_loader/qaic.py
```diff
@@ -147,6 +147,16 @@ def __init__(
         self.ctx_len = model_config.max_model_len
         self.decode_bsz = vllm_config.scheduler_config.max_num_seqs
         self.full_batch_size = vllm_config.scheduler_config.max_num_seqs
+        self.num_gpu_blocks_per_batch = self.ctx_len // self.prefill_seq_len
```
quic-vargupt, line 150: In case of P/D disaggregation, decode has a prefill_seq_len as 1 but for prefill the prefill_seq_len will be greater than 1, we end up having different page_memory and num_blocks on both P & D

quic-vargupt (changes requested):
Can you also try this PR in disagg mode?

vllm_qaic/model_loader/qaic.py
aditjadh: Agreed, I will make a single consolidated fix for P/D Disagg and MoE(prefill_seq_len=1)

vllm_qaic/platform_base.py
aditjadh: Ack

ankitqti:
Aditya add unit test

ankitqti:
Rebase PR and make PR on main branch

Also run precommit

aditjadh:
PR raised on main

PR #56, opened 2026-08-06
fix(qaic): Fix ccl edge cases in get_comp_ctx_lengths

quic-gthiruko:
Empty lists pass 'is not None' checks but cause IndexError/AssertionError downstream. Normalise to None so all guards behave correctly.

## Changes:
- Change `elif` to `if` for the decode shape check so a shape can match both prefill and decode buckets
- Drop the `ValueError` exception
- Return `None` instead of empty lists to ensure all guards behave correctly


ankitqti (approved):
Looks good

PR #59, opened 2026-08-07
Fix v0.23.0 mrotary_embedding_eager_fallback: is_neox_style=False

Neeharika-Gupta:
Fix v0.23.0 mrotary_embedding_eager_fallback configuration

### DIFF
--- a/vllm_qaic/ops/mrope.py
+++ b/vllm_qaic/ops/mrope.py
@@ -36,6 +36,15 @@ class QAicMRotaryEmbedding(MRotaryEmbedding):
         self._match_cos_sin_cache_dtype(query)
 
         # ------------------------------------------------------------------
+        # Interleaved rotation style (is_neox_style=False): the QAIC
+        # mrotary_embedding kernel and its eager fallback only support
+        # neox-style (first-half / second-half split). Fall back to the
+        # parent's pure-PyTorch implementation which handles both styles.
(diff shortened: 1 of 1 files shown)


vllm_qaic/ops/mrope.py
```diff
@@ -35,6 +35,15 @@ def forward_oot(
 
         self._match_cos_sin_cache_dtype(query)
 
+        # ------------------------------------------------------------------
+        # Interleaved rotation style (is_neox_style=False): the QAIC
+        # mrotary_embedding kernel and its eager fallback only support
+        # neox-style (first-half / second-half split). Fall back to the
+        # parent's pure-PyTorch implementation which handles both styles.
+        # ------------------------------------------------------------------
+        if not self.is_neox_style:
```
quic-agokhale: We can instead patch it so that we invoke the forward_oot() of the RoPE operator. check https://github.com/vllm-project/vllm/blob/0fc695fc6d1d82e9a5ac6835ac8e4e1c83703665/vllm/model_executor/layers/rotary_embedding/mrope.py#L305 where forward_native is called. We can have something similar, but instead invoke forward_oot() so that it routes to https://github.com/qualcomm/vllm-qaic/pull/84/changes#diff-3dab005ee5f580bb3df3128d3a65e4ecaadec5ce8d644f7c44f060708179756a

PR #61, opened 2026-08-07
MoE support v0.23.0

Shrinivas-kulkarni9:
Adding MoE support in v0.23.0 vllm. MoEs were broken due to upstream changes from v0.15.0 to v0.23.0 and these fixes address those issues


ankitqti (approved):
Looks good

PR #65, opened 2026-08-11
scripts: Make QEFF_BRANCH configurable via environment variable

quic-gthiruko:
Allow QEFF_BRANCH to be overridden via environment variable with default fallback to release/v1.22.0. This enables testing different qefficient branches without modifying utility.sh.


ankitqti:
Already part of another PR

PR #67, opened 2026-08-11
[v0.23.0][Enhancement] Add __main__ guard to qaic_eager_mode example

aditjadh:
Wrap the inference body of examples/qaic_eager_mode.py in a main() function guarded by:
 if __name__ == "__main__":

1. Previously all engine setup and generation ran at module level. Under the spawn multiprocessing start method, vLLM's engine-core child process re-imports the entry script, which would re-execute LLM() and llm.generate() in the child. The guard prevents this. 
2. Wrapping in main() also scopes llm/outputs as locals so device memory is released on return.

Brings the example in line with the pattern already used by qaic.py, qaic_spd.py, qaic_whisper.py, and the other examples. 
No functional change to the example's behavior.

### DIFF
--- a/examples/qaic_eager_mode.py
+++ b/examples/qaic_eager_mode.py
@@ -32,30 +32,36 @@ os.environ["QAIC_VISIBLE_DEVICES"] = (
     "0"  # for multiple devices, separate them through comma, i.e. 0,1,2,3
 )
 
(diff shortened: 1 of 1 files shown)


ankitqti (approved):
Looks good to me

PR #70, opened 2026-08-13
Fix Gemma4 image_embeds placeholder handling

quic-vargupt:
In vLLM v0.23, image_embeds inputs are validated as the logical image modality, but placeholder generation still calls Gemma4ForConditionalGeneration.get_placeholder_str() with the original modality name, image_embeds. Gemma4 only accepts image, audio, and video, which causes image_embeds requests to fail before multimodal processing.
This change maps image_embeds to image only for Gemma4 placeholder lookup, preserving the upstream Gemma4 image placeholder behavior while keeping the existing QAIC image_embeds multimodal processing path unchanged.


ankitqti (approved):
Looks good to me

PR #71, opened 2026-08-13
[DFlash] Add speculative decoding with Diffusion LM proposer

vjanfaza:
Adds DFlash speculative decoding support for Qualcomm QAIC devices. DFlash uses a Diffusion LM (DLM) as the draft model, conditioned on TLM hidden states captured after each prefill/decode step.

- vllm_qaic/spec_decode/dflash_draft_model.py: QaicDFlashProposer — CPU/numpy proposer driving the DLM QPC via np_run/complete_inf; handles prefill chunking (sub-block fan-out), decode proposal, and DLM KV synchronization via update_prefill_kv().
- vllm_qaic/model_loader/qaic.py: _derive_dflash_config() replicates _configure_dflash (sets _dflash_* on TLM/DLM hf_configs); compile config wires prefill_only/num_speculative_tokens/target_layer_ids; get_hf_model() unwraps EAGLEConfig for the DLM draft branch.
- vllm_qaic/worker/model_runner.py: DFlash runner wiring — method allowlist, proposer instantiation/load, per-step hidden-state capture (prefill chunks + decode buffer), propose dispatch, and unconditional update_prefill_kv() after bookkeeping.

## Equally Contributors
- @vjanfaza
- @fannanya

### DIFF
--- a/docs/docs/user_guide/features/index.md
+++ b/docs/docs/user_guide/features/index.md
@@ -25,7 +25,7 @@ Status of vLLM features on Qualcomm Cloud AI hardware.
 | N-gram | ✅ | ❌ | |
 | Suffix | ✅ | ❌ | |
 | Draft model | ✅ | ❌ | Separate DLM on same device |
-| DFlash (block-diffusion draft) | ✅ | ❌ | Separate DLM on same device |
+| DFlash (block-diffusion draft) | ✅ | ❌ | Separate DLM |
 | **Advanced Features** | | | |
 | LoRA adapters | ✅ | ❌ | Hot-swap adapters |
 | Disaggregated serving | ✅ | ❌ | xEyPzD prefill/decode split |
--- a/docs/docs/user_guide/features/index.md
+++ b/docs/docs/user_guide/features/index.md
@@ -25,6 +25,7 @@ Status of vLLM features on Qualcomm Cloud AI hardware.
 | N-gram | ✅ | ❌ | |
 | Suffix | ✅ | ❌ | |
 | Draft model | ✅ | ❌ | Separate DLM on same device |
+| DFlash (block-diffusion draft) | ✅ | ❌ | Separate DLM on same device |
 | **Advanced Features** | | | |
 | LoRA adapters | ✅ | ❌ | Hot-swap adapters |
 | Disaggregated serving | ✅ | ❌ | xEyPzD prefill/decode split |
--- a/docs/docs/user_guide/features/speculative_decoding.md
+++ b/docs/docs/user_guide/features/speculative_decoding.md
@@ -16,13 +16,14 @@ Speculative decoding (SpD) accelerates token generation by using a fast proposer
 
 ## Methods
 
-Cloud AI supports three speculative decoding methods:
+Cloud AI supports four speculative decoding methods:
 
 | Method | Proposer | Separate Model Binary | Dedicated Core Allocation | Best For |
 |--------|----------|-------------|-------------|----------|
 | `ngram` | N-gram pattern matching | No | No | Summarization, repetitive output |
 | `suffix` | Suffix array matching | No | No | Code completion, long-context echo |
 | `draft_model` | Lightweight LLM | Yes | Yes | General text generation |
+| `dflash` | Block-diffusion draft LM | Yes | Yes | High-throughput block drafting |
 
 ## N-gram Speculative Decoding
 
@@ -117,6 +118,80 @@ docker run --rm -it --network host \
     Draft model compilation requires separate QPC generation. See the
     [QEfficient SpD Guide](https://quic.github.io/efficient-transformers/speculative_decoding.html)
     for compilation details.
+
+## DFlash Speculative Decoding
+
+DFlash uses a block-diffusion draft LM (DLM) that proposes an entire block of candidate tokens per step in a single batched forward pass, conditioned on the target model's (TLM) hidden states. The DLM's `block_size` sets `num_speculative_tokens`:
+
+```python
+from vllm import LLM, SamplingParams
+
+device_group = [0, 1, 2, 3]
+block_size = 16  # DFlash DLM block_size == num_speculative_tokens
+
+llm = LLM(
+    model="Qwen/Qwen3-4B",
+    max_num_seqs=4,
+    max_model_len=4096,
+    long_prefill_token_threshold=128,  # TLM prefill_seq_len (multiple of block_size)
+    quantization="mxfp6",
+    kv_cache_dtype="mxint8",
+    additional_config={
+        "override_qaic_config": {
+            "device_group": device_group,
+            "num_cores": 8,
+            "prefill_seq_len": 128,
+            "mxfp6_matmul": True,
+            "mxint8_kv_cache": True,
+            "mos": 1,
+        },
+        "draft_override_qaic_config": {
+            "device_group": device_group,  # Same device
+            "num_cores": 8,
+            "prefill_seq_len": block_size,  # DLM prefills block_size at a time
+            "mxfp6_matmul": True,
+            "mxint8_kv_cache": True,
+            "mos": 1,
+        },
+    },
+    speculative_config={
+        "method": "dflash",
+        "model": "z-lab/Qwen3-4B-DFlash-b16",
+        "num_speculative_tokens": block_size,
+    },
+)
+```
+
+**Docker example:**
+
+```bash
+docker run --rm -it --network host \
+  --device /dev/accel/ \
+  --shm-size=4gb \
+  -e HF_TOKEN=<your_hf_token> \
+  ghcr.io/quic/cloud_ai_inference_vllm:1.21.2.0 \
+  --host 127.0.0.1 --port 8000 \
+  --model Qwen/Qwen3-4B \
+  --max-model-len 4096 \
+  --max-num-seq 4 \
+  --max-seq-len-to-capture 128 \
+  --quantization mxfp6 \
+  --kv-cache-dtype mxint8 \
+  --speculative-config '{"method":"dflash","model":"z-lab/Qwen3-4B-DFlash-b16","num_speculative_tokens":16}' \
+  --additional-config '{"override_qaic_config":{"device_group":[0,1,2,3],"num_cores":8,"prefill_seq_len":128,"mxfp6_matmul":true,"mxint8_kv_cache":true,"mos":1},"draft_override_qaic_config":{"device_group":[0,1,2,3],"num_cores":8,"prefill_seq_len":16,"mxfp6_matmul":true,"mxint8_kv_cache":true,"mos":1}}'
+```
+
+!!! info "QEfficient Reference"
+    The DFlash DLM checkpoint (e.g. `z-lab/Qwen3-4B-DFlash-b16`) and its
+    cross-checkpoint config (`block_size`, `target_layer_ids`, `mask_token_id`)
+    are consumed via QEfficient's `DFlashDLMTransform`/`DFlashTLMTransform`. See
+    the [QEfficient SpD Guide](https://quic.github.io/efficient-transformers/speculative_decoding.html)
+    for details.
+
+!!! note "TLM `prefill_seq_len` must be a multiple of `block_size`"
+    The target model's `prefill_seq_len` (`long_prefill_token_threshold`) must be
+    an exact multiple of the DLM's `block_size`, since prefill chunks are fanned
+    out into `block_size`-sized sub-blocks for the DLM.
 
 ## Core Allocation
 
--- a/examples/qaic_spd.py
+++ b/examples/qaic_spd.py
@@ -61,7 +61,7 @@ def main() -> None:
             },
         },
         speculative_config={
-            "method": "draft_model",
+            "method": "draft_model",  # for DFlash use "dflash" and point model to a DFlash DLM checkpoint
             "model": "meta-llama/Llama-3.2-1B-Instruct",
             "num_speculative_tokens": 3,
         },
--- a/examples/qaic_spd.py
+++ b/examples/qaic_spd.py
@@ -61,7 +61,7 @@ def main() -> None:
             },
         },
         speculative_config={
-            "method": "draft_model",  # for DFlash use "dflash" and point model to a DFlash DLM checkpoint
+            "method": "draft_model",
             "model": "meta-llama/Llama-3.2-1B-Instruct",
             "num_speculative_tokens": 3,
         },
--- a/examples/qaic_spd_dflash.py
+++ b/examples/qaic_spd_dflash.py
@@ -5,9 +5,6 @@
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 # SPDX-License-Identifier: Apache-2.0
 # Adapted from examples/qaic_spd.py — DFlash (Diffusion-LM draft) SpD.
-#
-# Local dev/test script (NOT part of the PR). vLLM logs spec-decode acceptance
-# natively when disable_log_stats=False.
 
 import gc
 import random
--- a/examples/qaic_spd_dflash.py
+++ b/examples/qaic_spd_dflash.py
@@ -7,9 +7,22 @@
 # Adapted from examples/qaic_spd.py — DFlash (Diffusion-LM draft) SpD.
 
 import gc
+import json
 import random
 
+from huggingface_hub import hf_hub_download
 from vllm import LLM, SamplingParams
+
+
+def _dflash_block_size(dlm_repo: str) -> int:
+    """DLM block_size (== num_speculative_tokens) read from the checkpoint config."""
+    cfg_path = hf_hub_download(repo_id=dlm_repo, filename="config.json")
+    with open(cfg_path) as f:
+        cfg = json.load(f)
+    bs = cfg.get("block_size") or cfg.get("dflash_config", {}).get("block_size")
+    if bs is None:
+        raise ValueError(f"no block_size in {dlm_repo}/config.json")
+    return int(bs)
 
 
 def main() -> None:
@@ -26,13 +39,14 @@ def main() -> None:
     # QPC parameters
     ctx_len = 4096
     seq_len = 128  # TLM prefill_seq_len (multiple of block_size)
-    block_size = 16  # DFlash DLM block_size == num_speculative_tokens
     decode_bsz = 4
 
     device_group = [0, 1, 2, 3]
 
     tlm_repo = "Qwen/Qwen3-4B"
     dlm_repo = "z-lab/Qwen3-4B-DFlash-b16"
+    # block_size (== num_speculative_tokens) is a property of the DLM checkpoint.
+    block_size = _dflash_block_size(dlm_repo)
 
     print(
         f"DFlash SpD run (TLM={tlm_repo} -> DLM={dlm_repo}, "
--- a/examples/qaic_spd_dflash.py
+++ b/examples/qaic_spd_dflash.py
@@ -1,0 +1,97 @@
+# ------------------------------------------------------------------
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+# ------------------------------------------------------------------
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+# SPDX-License-Identifier: Apache-2.0
+# Adapted from examples/qaic_spd.py — DFlash (Diffusion-LM draft) SpD.
+#
+# Local dev/test script (NOT part of the PR). vLLM logs spec-decode acceptance
+# natively when disable_log_stats=False.
+
+import gc
+import random
+
+from vllm import LLM, SamplingParams
+
+
+def main() -> None:
+    prompts = [
+        "The history of artificial intelligence dates back",
+        "Speculative decoding accelerates inference by",
+        "Large language models are trained on",
+        "The key difference between supervised and unsupervised learning is",
+    ] * 5
+    random.shuffle(prompts)
+
+    sampling_params = SamplingParams(temperature=0.0, max_tokens=200)
+
+    # QPC parameters
+    ctx_len = 4096
+    seq_len = 128  # TLM prefill_seq_len (multiple of block_size)
+    block_size = 16  # DFlash DLM block_size == num_speculative_tokens
+    decode_bsz = 4
+
+    device_group = [0, 1, 2, 3]
+
+    tlm_repo = "Qwen/Qwen3-4B"
+    dlm_repo = "z-lab/Qwen3-4B-DFlash-b16"
+
+    print(
+        f"DFlash SpD run (TLM={tlm_repo} -> DLM={dlm_repo}, "
+        f"block_size={block_size})...\n"
+    )
+
+    llm = LLM(
+        model=tlm_repo,
+        max_num_seqs=decode_bsz,
+        max_model_len=ctx_len,
+        long_prefill_token_threshold=seq_len,
+        quantization="mxfp6",
+        kv_cache_dtype="mxint8",
+        disable_log_stats=False,  # native spec-decode acceptance metrics
+        gpu_memory_utilization=1.0,
+        async_scheduling=False,
+        trust_remote_code=True,
+        additional_config={
+            "override_qaic_config": {
+                "device_group": device_group,
+                "num_cores": 8,
+                "prefill_seq_len": seq_len,
+                "mxfp6_matmul": True,
+                "mxint8_kv_cache": True,
+                "mos": 1,
+            },
+            "draft_override_qaic_config": {
+                "device_group": device_group,
+                "num_cores": 8,
+                "prefill_seq_len": block_size,  # DLM prefills block_size at a time
+                "mxfp6_matmul": True,
+                "mxint8_kv_cache": True,
+                "mos": 1,
+            },
+        },
+        speculative_config={
+            "method": "dflash",
+            "model": dlm_repo,
+            "num_speculative_tokens": block_size,
+        },
+    )
+
+    outputs = llm.generate(prompts, sampling_params)
+
+    for output in outputs:
+        prompt = output.prompt
+        generated_text = output.outputs[0].text
+        num_generated_tokens = len(output.outputs[0].token_ids)
+        print(
+            f"Prompt: {prompt!r}, Generated text: {generated_text!r}, "
+            f"Num generated tokens: {num_generated_tokens!r}"
+        )
+
+    del llm
+    gc.collect()
+
+
+if __name__ == "__main__":
+    main()
--- a/tests/e2e/test_qaic_dflash.py
+++ b/tests/e2e/test_qaic_dflash.py
@@ -107,3 +107,56 @@ class TestQaicDFlash:
             assert len(set(generations)) == 1, (
                 "Outputs from different slots for same prompt does not match!!"
             )
+
+
+# Not part of TestQaicDFlash: this stands up two models itself (base, then
+# base+DFlash) via qaic_runner_factory, so it does not use the class's single
+# dflash `qaic_model`. The marker only advertises the device count (4) so the
+# collection-time skip logic reserves enough devices.
+@pytest.mark.qaic_test_config(num_device_groups=1, device_group_size=4)
+def test_dflash_matches_base_greedy(qaic_runner_factory, sample_prompts):
+    """Lossless-equivalence check: greedy decoding from the base TLM alone must be
+    token-identical to greedy decoding from base + DFlash. DFlash only accepts a
+    draft when it equals the TLM's own argmax, so the sequences must match exactly.
+
+    Full precision (no mxfp6/mxint8 — compiler default 16-bit) keeps the TLM
+    numerically identical across both runs, so an exact match is the right bar.
+    The two models are built sequentially to hold only 4 devices at a time."""
+    prompts = list(sample_prompts)[:4]
+    max_tokens = 32
+    tlm = "Qwen/Qwen3-4B"
+    # Full-precision compile: no quantization knobs (mxfp6/mxint8 omitted).
+    tlm_override = {"num_cores": 8, "prefill_seq_len": _DFLASH_SEQ_LEN}
+
+    with qaic_runner_factory(
+        tlm,
+        speculative_config=None,
+        seq_len=_DFLASH_SEQ_LEN,
+        override_qaic_config=tlm_override,
+        device_group_size=4,
+    ) as base_model:
+        base = base_model.generate_greedy(prompts, max_tokens)
+
+    with qaic_runner_factory(
+        tlm,
+        speculative_config={
+            "method": "dflash",
+            "model": "z-lab/Qwen3-4B-DFlash-b16",
+            "num_speculative_tokens": _DFLASH_BLOCK_SIZE,
+        },
+        seq_len=_DFLASH_SEQ_LEN,
+        override_qaic_config=tlm_override,
+        draft_override_qaic_config={
+            "num_cores": 8,
+            "prefill_seq_len": _DFLASH_BLOCK_SIZE,
+        },
+        device_group_size=4,
+    ) as dflash_model:
+        spec = dflash_model.generate_greedy(prompts, max_tokens)
+
+    for i, ((base_ids, _), (spec_ids, _)) in enumerate(zip(base, spec, strict=True)):
+        assert spec_ids == base_ids, (
+            f"DFlash greedy output diverged from base greedy for prompt {i}:\n"
+            f"  base  : {base_ids}\n"
+            f"  dflash: {spec_ids}"
+        )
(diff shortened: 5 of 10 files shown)


examples/qaic_spd.py
```diff
@@ -61,7 +61,7 @@ def main() -> None:
             },
         },
         speculative_config={
-            "method": "draft_model",
+            "method": "draft_model",  # for DFlash use "dflash" and point model to a DFlash DLM checkpoint
```
ankitqti: Make a separate example file for dflash something as qaic_spd_dflash.py

vllm_qaic/worker/model_runner.py
```diff
@@ -483,6 +487,36 @@ def __init__(
                     },
                 )
             self.drafter = QaicDraftModelProposer(draft_vllm_config)
+        elif (
```
ankitqti: Why do we merge this elsif condition with prevision if condition

vllm_qaic/worker/model_runner.py
```diff
@@ -483,6 +487,36 @@ def __init__(
                     },
                 )
             self.drafter = QaicDraftModelProposer(draft_vllm_config)
```
ankitqti: Merge both if conditions

self.drafter = QaicDraftModelProposer(draft_vllm_config) if self.speculative_config.uses_draft_model() else QaicDFlashProposer(draft_vllm_config)

vllm_qaic/worker/model_runner.py
```diff
@@ -1566,6 +1673,15 @@ def load_model(self, *args, **kwargs) -> None:
             and self.drafter is not None
         ):
             self.drafter.load_model()
+        elif (
+            self.speculative_config is not None
+            and self.speculative_config.use_dflash()
+            and self.drafter is not None
+        ):
+            # DFlash: the DLM's sub-block fan-out is sized by the TLM prefill_seq_len.
+            self.drafter.tlm_prefill_seq_len = self.model.prefill_seq_len
```
ankitqti, line 1690: why drafter prefill seq len will be different than model seq_len always ?

ankitqti (commented):
Some comments

vllm_qaic/worker/model_runner.py
vjanfaza: The TLM and DLM intentionally use different prefill_seq_len values to generalize it for TTFT of TLM — it's required to keep the DLM to a single compilation specialization.
DFlash's DLM operates on fixed-size blocks (block_size, e.g. 8 or 16): during decode it always takes block_size masked/noise tokens and emits block_size draft tokens in one shot. We reuse that same block_size as the DLM's prefill chunk size, so the DLM only ever sees one input shape — block_size — whether it's being fed a prefill chunk or a decode step. That lets it compile with a single specialization (prefill_only=True, prefill_seq_len=block_size).
The TLM, on the other hand, prefills the prompt at a larger prefill_seq_len (e.g. 128) for prefill throughput, independent of block_size.
We tried unifying both to the same prefill_seq_len early on. That forced the DLM to compile two specializations (one sized for prefill, one for decode), and end-to-end decode throughput dropped by roughly half compared to the single-specialization version. Given that regression, keeping tlm_prefill_seq_len (read from the loaded TLM, self.model.prefill_seq_len) separate from the DLM's own block_size-sized prefill is deliberate — dflash_draft_model.py asserts tlm_prefill_seq_len % block_size == 0 and chunks the TLM's prefill into block_size-sized sub-blocks for the DLM accordingly.

vllm_qaic/worker/model_runner.py
vjanfaza: Fixed!

vllm_qaic/worker/model_runner.py
vjanfaza: Fixed!

examples/qaic_spd.py
vjanfaza: Done!

eplatero97: let's leave this file unmodified

examples/qaic_spd_dflash.py
```diff
@@ -0,0 +1,97 @@
+# ------------------------------------------------------------------
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+# ------------------------------------------------------------------
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+# SPDX-License-Identifier: Apache-2.0
+# Adapted from examples/qaic_spd.py — DFlash (Diffusion-LM draft) SpD.
+#
+# Local dev/test script (NOT part of the PR). vLLM logs spec-decode acceptance
+# natively when disable_log_stats=False.
```
eplatero97: this file is part of PR, right? if so, let's remove these comments

vllm_qaic/worker/model_runner.py
```diff
@@ -1240,6 +1266,20 @@ def execute_model(
...
+                    _hidden_size = self.model_config.get_hidden_size()
+                    _pfl = self.model.prefill_seq_len
+                    # Must match the QPC's hidden_states binding dtype (float16 for
+                    # mxfp6/mxint8 models) or setData rejects the buffer.
+                    _hs_dtype = self._dflash_hidden_dtype
+                    _dflash_hidden_chunks = [
+                        np.zeros((1, _pfl, _hidden_size), dtype=_hs_dtype)
+                        for _ in range(_n_chunks)
+                    ]
+                    self.model._dflash_hidden_state_chunks = _dflash_hidden_chunks
```
eplatero97: instead of passing this through instance variable, why don't we just pass it to `self.mode(.)`?

ankitqti:
Any performance numbers you have with this PR ?

eplatero97:
PR looks mostly good (left some comments). 

We also need to update `docs/` directory to advertise that D-Flash is now supported.

Critical next step is to create unit tests that assert D-flash functionality under different scenarios (e.g., d-flash works as expected when we have prefill chunks, decode-prefill inputs in one batch). 

Lastly, we need to assert all existing unit tests are passing to preserve backward compatibility

ankitqti:
Can we see if we can use vllm public repo Dflash unit tests ?

vjanfaza:
Thanks for the review! Addressed all three points:

  Docs: Added DFlash to the features table (docs/user_guide/features/index.md) and a full usage section in speculative_decoding.md (config example, Docker CLI, prefill_seq_len/block_size note).

  New tests (tests/e2e/test_qaic_dflash.py):
  - test_prefill_chunking — prompts forced (via seeded ShareGPT sampling, min_len > prefill_seq_len) to exceed the TLM's chunk size, exercising multi-chunk prefill with hidden-state capture per chunk.
  - test_mixed_decode_prefill_batch — more requests than decode_bsz, forcing continuous batching to schedule new prefills alongside in-flight decodes in the same step.
  - test_generate / test_output_consistency — baseline correctness and determinism (same outputs across shuffled batch-slot ordering).
test_prefill_chunking actually caught a real bug: building the DFlash draft model's VllmConfig reused the target's scheduler_config by reference, so check_and_update_config recomputed the scheduling budget using the DLM's tiny prefill_seq_len and leaked that shrunk value back into the target's live scheduler — permanently stalling any prefill chunk larger than the shrunk budget. Fixed by deep-copying scheduler_config before constructing the draft config, scoped to spec_cfg.use_dflash() only.
Backward compatibility: All 4 new DFlash tests pass. The fix is gated behind use_dflash() so it's unreachable for other speculative methods; the conftest.py sampling helper changes (seed/min_len params) are additive with defaults matching prior behavior, verified against all existing callers. Full test suite re-run confirms no regressions outside this path.

vjanfaza:
> Any performance numbers you have with this PR ?

Working on them, Ankit. we have numbers for v0.15.0 in the following link and will publish numbers for v0.23.0 as well. Also, we are working on gemma4 and qwen3-vl models and the following link will get updated accordingly.
[DFlash-VLLM-Results](https://qualcomm-my.sharepoint.com/:x:/r/personal/fannanya_qti_qualcomm_com1/_layouts/15/Doc.aspx?sourcedoc=%7B9B4C2503-0C2E-49A9-8EE1-1CFFAFCAE713%7D&file=[summary_model_dflash_v1.xlsx](https://qualcomm-my.sharepoint.com/:x:/r/personal/fannanya_qti_qualcomm_com1/_layouts/15/Doc.aspx?sourcedoc=%7B9B4C2503-0C2E-49A9-8EE1-1CFFAFCAE713%7D&file=summary_model_dflash_v1.xlsx&action=default&mobileredirect=true)&action=default&mobileredirect=true)

vllm_qaic/worker/model_runner.py
```diff
@@ -1522,12 +1594,35 @@ def propose_draft_token_ids(
...
+        the TLM decode hidden states into it via the bound output."""
+        block_size = self.speculative_config.num_speculative_tokens
+        hidden_size = self.model_config.get_hidden_size()
+        decode_bsz = self.model.decode_bsz
+        _hs_info = self.model.get_io_shape_and_dtype("hidden_states", is_input=False)
+        self._dflash_hidden_dtype = _hs_info[1] if _hs_info is not None else np.float32
+        self._tlm_hidden_buf = np.zeros(
+            (decode_bsz, block_size, hidden_size), dtype=self._dflash_hidden_dtype
+        )
+        self.model._dflash_decode_hidden_buf = self._tlm_hidden_buf
```
eplatero97: Can we instead pass this as input to `self.model` similar to how we did with `_dflash_hidden_chunks`? 

also, suppose that one request exceeds `max_model_len` when proposing tokens. This will disable proposing drafts for that step. In this scenario, if we have decode requests that went through verification, we still need to be able to update DLM with the TLM decode buffers, right? If so, is this handled?

vllm_qaic/worker/model_runner.py
```diff
@@ -1240,6 +1274,19 @@ def execute_model(
...
+                    _n_chunks = self.model.num_prefill_chunks(prefill_cum_sum)
+                    _hidden_size = self.model_config.get_hidden_size()
+                    _pfl = self.model.prefill_seq_len
+                    # Must match the QPC's hidden_states binding dtype (float16 for
+                    # mxfp6/mxint8 models) or setData rejects the buffer.
+                    _hs_dtype = self._dflash_hidden_dtype
+                    _dflash_hidden_chunks = [
+                        np.zeros((1, _pfl, _hidden_size), dtype=_hs_dtype)
+                        for _ in range(_n_chunks)
+                    ]
```
eplatero97: **optimization:** would it be worth it to cache `_dflash_hidden_chunks` as an instance variable to avoid remaking numpy buffers?  

also, variable `_dflash_hidden_chunks` seems to imply these are dflash hidden chunks. maybe we can make it more general to something like `tlm_prefill_hidden_chunks`? this way we can also reuse it for other techniques like EAGLE?

vllm_qaic/worker/model_runner.py
eplatero97: I do recall an earlier discussion where we determined that creating two prefill specializations for D-Flash resulted in a significant performance regression.

A related optimization would be to avoid emitting block_size logits during prefill. Instead, we could return a single logit per request, which would reduce both compute requirements and DMA bandwidth.

My assumption is that this approach would suffer from a similar performance degradation due to the same compiler issue. If that's the case, this compiler bug is already blocking at least two potential optimizations, which may make it worthwhile to raise with the broader team.

examples/qaic_spd_dflash.py
```diff
@@ -0,0 +1,94 @@
...
+        "The key difference between supervised and unsupervised learning is",
+    ] * 5
+    random.shuffle(prompts)
+
+    sampling_params = SamplingParams(temperature=0.0, max_tokens=200)
+
+    # QPC parameters
+    ctx_len = 4096
+    seq_len = 128  # TLM prefill_seq_len (multiple of block_size)
+    block_size = 16  # DFlash DLM block_size == num_speculative_tokens
```
eplatero97: https://huggingface.co/z-lab/Qwen3-4B-DFlash-b16

On above link, they show below snippet:

```bash
vllm serve Qwen/Qwen3-4B \
  --speculative-config '{"method": "dflash", "model": "z-lab/Qwen3-4B-DFlash-b16", "num_speculative_tokens": 15}' \
  --attention-backend flash_attn \
  --max-num-batched-tokens 32768
```

in it, they are passing `num_speculative_tokens = block_size - 1`. should we do the same to align with API? 

also, if we don't provide `num_speculative_tokens` similar to how they show with below sglang block, does vllm automatically recognize `block_size` or do we have to provide it explicitly?

```bash
# Optional: enable schedule overlapping (experimental, may not be stable)
# export SGLANG_ENABLE_SPEC_V2=1
# export SGLANG_ENABLE_DFLASH_SPEC_V2=1
# export SGLANG_ENABLE_OVERLAP_PLAN_STREAM=1

python -m sglang.launch_server \
    --model-path Qwen/Qwen3-4B \
    --speculative-algorithm DFLASH \
    --speculative-draft-model-path z-lab/Qwen3-4B-DFlash-b16 \
    --tp-size 1 \
    --dtype bfloat16 \
    --attention-backend fa3 \
    --mem-fraction-static 0.75 \
    --trust-remote-code
```

docs/docs/user_guide/features/index.md
```diff
@@ -25,6 +25,7 @@ Status of vLLM features on Qualcomm Cloud AI hardware.
 | N-gram | ✅ | ❌ | |
 | Suffix | ✅ | ❌ | |
 | Draft model | ✅ | ❌ | Separate DLM on same device |
+| DFlash (block-diffusion draft) | ✅ | ❌ | Separate DLM on same device |
```
eplatero97: `Separate DLM on same device` is not always true (we can have draft run on its own device). 

this was a mistake on my side so apologies on that. 

can we fix it to say something like `Separate DLM`?

eplatero97 (commented):
few potential changes to look at in regard of functionality, variable names, and client-API

vllm_qaic/worker/model_runner.py
vjanfaza: Good catch — this was a real gap. When one request exceeded effective_drafter_max_model_len, the batch-wide input_fits_in_drafter gate skipped propose() entirely, so every other active decode request's DLM KV cache (position_counter) silently fell behind the TLM.
Fixed in ba4fe91: added a commit: bool = True param to QaicDFlashProposer.propose(). When the gate is closed, the runner now calls propose(..., commit=False), which still runs the DLM forward to advance KV/position_counter for all active requests but discards the resulting draft candidates instead of offering them. Added tests/unit/test_dflash_kv_advance.py to cover this directly (KV advances under commit=False, no candidates are offered/cached, and a later commit=True call stays correctly in sync). The change is scoped entirely to the DFlash code path (propose()'s new param, gated behind speculative_config.use_dflash() in the runner), so other speculative-decoding methods (ngram, etc.) and non-spec-decode paths are unaffected.

ankitqti:
Move this PR to main

vjanfaza:
> Move this PR to main

Hi Ankit. I moved it to main branch.

examples/qaic_spd_dflash.py
vjanfaza: For your first question, I prefer to keep it as it is since in the implementation we have similar implementation as public vllm during runtime but for us block_size in the example file, is consumed directly as the DLM's compile-time block_size (buffer shapes, DLM prefill_seq_len, num_logits_to_keep, sub-block tiling) and is shared with efficient-transformers' compilation — it's a compile hyperparameter, not a runtime draft-length. 

I updated qaic_spd_dflash.py to read block_size directly from  DLM config.json. Similar implementation exist in QEfficient.

vllm_qaic/spec_decode/dflash_draft_model.py
```diff
@@ -0,0 +1,414 @@
...
+                        [req_n_tokens], dtype=prefill_cum_sum.dtype
+                    ),
+                    "batch_indices": req_block_ids,
+                    "prefill_is_partial": prefill_is_partial[req_i : req_i + 1],
+                    "last_chunk_logits": req_logits,
+                    "req_id": req_id,
+                }
+            )
+            chunk_idx += 1
+            tok_start = int(tok_end)
```
eplatero97, line 195: it seems to me like this is doing one chunk per request. if so, shouldn't this be doing `ceil(seq_len/dlm_block_size)` chunks?

vllm_qaic/spec_decode/dflash_draft_model.py
```diff
@@ -0,0 +1,414 @@
...
+            if not accepted_seq:
+                continue
+
+            if st.candidates_from_prefill:
+                # Skip forward; keep cached candidates unless committing.
+                if commit:
+                    st.candidates_from_prefill = False
+                    assert st.dlm_candidates is not None
+                    draft_token_ids[i] = st.dlm_candidates[1:].tolist()
+                continue
```
eplatero97: Possible DLM desync: a freshly prefilling request appears to be frozen across sequential closed-gate `commit=False` steps.

my understanding is that when `commit=False`, prefill steps are intentionally skipped so the position counter is not updated. That makes sense. However, because `candidates_from_prefill = True`, it looks like the next request may also skip proposing. During `propose(..., commit=False)`, the `candidates_from_prefill` branch skips the 1-token decode update as well.

for example, for a request R with prompt length P:
```text
Step T (prefill, gate closed)
- position_counter = P - 1
- candidates are cached
- execution continues
```

This is correct.

```text
Step T+1 (gate still closed)
- The TLM confirms position P + 1
- R hits the candidates_from_prefill path and continues again
- position_counter remains pinned at P - 1
- sibling requests continue advancing normally
```
```text
Gate reopens at T+k
- propose(commit=True) serves the cached candidates
- those candidates are now k positions stale
- the decode forward computes:

      row_pids_target = arange(prev_counter + 1, ...)

  using:

      prev_counter = P - 1
```

this appears misaligned with the actual sequence already confirmed by the TLM

vllm_qaic/spec_decode/dflash_draft_model.py
```diff
@@ -0,0 +1,414 @@
...
+    def update_prefill_kv(self) -> None:
+        """Drain pending DLM prefill work items; no-op when nothing pending."""
+        for ctx in self._prefill_pending:
+            self.prefill_step(**ctx)
+        self._prefill_pending = []
+
+    def prefill_step(
+        self,
+        target_hidden: np.ndarray,
+        prefill_input_ids: np.ndarray,
```
eplatero97: I see that `build_prefill_pending` always sets `prefill_input_ids = None`

eplatero97 (commented):
left some comments that may require additional changes. if found that functional changes are needed, let's make sure to create unit tests that cover gaps. 

lastly, let's create a unit test that asserts outputs of base model are the same as outputs of (base + dflash)

vllm_qaic/spec_decode/dflash_draft_model.py
vjanfaza: This one's fine as-is. build_prefill_pending just creates one work item per request per TLM prefill chunk — it's not where the block_size splitting happens. The ceil(n_tokens / block_size) sub-block loop lives inside prefill_step, which is where each chunk's hidden states actually get fed to the DLM. So the item count here is correct. For making sure about this I have tried benchmarking with smaller prefill_seq_len to fall into more chunks as well and it works perfectly.

vllm_qaic/spec_decode/dflash_draft_model.py
vjanfaza: Good catch, this was a real bug. When the gate is closed we call propose(commit=False), and a request that just finished prefill (candidates_from_prefill=True) was hitting the continue and skipping the DLM forward entirely. so its position_counter stayed pinned while the TLM kept moving, and the DLM KV drifted out of sync. Fixed it to drop the cached prefill drafts (they can't be served on a closed gate anyway) and fall through to the normal KV-advance path. Added a unit test for both the closed- and open-gate cases.

vllm_qaic/spec_decode/dflash_draft_model.py
vjanfaza: Yep, it's dead. prefill_step never reads it — the DLM prefill runs off mask tokens plus the TLM hidden states, not raw input ids. Removed the dict key and the parameter.

vjanfaza:
> left some comments that may require additional changes. if found that functional changes are needed, let's make sure to create unit tests that cover gaps.
> 
> lastly, let's create a unit test that asserts outputs of base model are the same as outputs of (base + dflash)

Done. Added the unit tests above for the closed-gate fix, plus an e2e test (test_dflash_matches_base_greedy) that runs greedy decode on the base model alone and on base+DFlash and checks the token ids match exactly. It compiles full precision (no mxfp6/mxint8) so the TLM is numerically identical across both runs, which makes an exact match the right bar. A small factory fixture builds the two models one after the other, so it only holds 4 devices at a time.

examples/qaic_spd_dflash.py
eplatero97: I agree that internally, our implementation is similar. However, I'm having a hard time justifying why our *public* API for d-flash is different than all other backends who take `num_speculative_tokens=block_size-1`. I think this will cause confusion. 

Also, in our implementation, I see code like below:
```python
        # DFlash TLM compiles with K-1 spec tokens (bonus token in slot 0).
        if _method == "dflash" and self.num_spec_tokens > 0:
            self.decode_ks = [self.num_spec_tokens - 1]
```

instead, we could pass `self.num_spec_tokens+1` to QEff and keep public API consistent

vllm_qaic/spec_decode/dflash_draft_model.py
eplatero97: You're right. In reality, I think this only works because currently, vLLM schedules one chunk per request. 

However, [PR 74](https://github.com/qualcomm/vllm-qaic/pull/74) will make changes that will allow >1 chunk per request. Once merged, this will break. 

I think we should generalize this to handle multiple chunks per request so that we don't have to re-raise future PR to solve this but let me know what you think

eplatero97 (commented):
well done, this is looking good!

I left some final comments.

can you also please post link to QEff d-flash PR?

eplatero97:
also, do we have perf numbers of d-flash with vllm implementation? ideally in comparison to vanilla or other speculative decoding methods?

vllm_qaic/spec_decode/dflash_draft_model.py
vjanfaza: Good call — you're right that this only worked because vLLM currently schedules one chunk per request, and PR #74 breaks that assumption. I've generalized both build_prefill_pending (chunk cursor now advances by ceil(req_tokens / prefill_seq_len), and each request gets its own run of hidden-state buffers) and prefill_step (accepts a list of per-chunk buffers; sub-blocks route to the right chunk — safe since prefill_seq_len is a multiple of block_size, so no block straddles a boundary). Added 3 unit tests covering the multi-chunk fan-out and the cross-request buffer alignment. PR #74 is config-layer-only (moves the chunk-length source to override_qaic_config["prefill_seq_len"]) and doesn't touch this path, so there's no conflict — this fix just consumes what #74 will produce.

examples/qaic_spd_dflash.py
vjanfaza: Fixed. I flipped the public API to num_speculative_tokens = block_size - 1 so it matches every other vLLM spec-decode backend and the model-card example.

vllm_qaic/worker/model_runner.py
```diff
@@ -520,8 +520,9 @@ def __init__(
         self.num_decode_tokens = 0
         self.max_decode_tokens = 1 + self.num_spec_tokens
         # DFlash emits block_size logits/step (not 1+K); must match QaicCausalLM.
+        # Public num_speculative_tokens is block_size - 1, so block_size = K + 1.
         if self.speculative_config and self.speculative_config.method == "dflash":
-            self.max_decode_tokens = self.num_spec_tokens
+            self.max_decode_tokens = self.num_spec_tokens + 1
```
eplatero97, line 525: isn't line 521 already doing this?

vllm_qaic/worker/model_runner.py
```diff
@@ -530,9 +531,10 @@ def __init__(
             if _method in ("ngram", "suffix") and self.max_decode_tokens > 1
             else [self.num_spec_tokens]
         )
-        # DFlash TLM compiles with K-1 spec tokens (bonus token in slot 0).
+        # DFlash public num_speculative_tokens already excludes the slot-0 bonus
+        # token, so the TLM decodes exactly that many spec tokens.
         if _method == "dflash" and self.num_spec_tokens > 0:
-            self.decode_ks = [self.num_spec_tokens - 1]
+            self.decode_ks = [self.num_spec_tokens]
```
eplatero97, line 537: I think we can merge this with line 529-532

vllm_qaic/worker/model_runner.py
```diff
@@ -1628,7 +1630,8 @@ def _setup_dflash_hidden_capture(self) -> None:
         """Allocate + bind the TLM decode hidden-state buffer for DFlash. The DLM
         proposer reads self._tlm_hidden_buf each decode step; _run_decode captures
         the TLM decode hidden states into it via the bound output."""
-        block_size = self.speculative_config.num_speculative_tokens
+        # Public num_speculative_tokens is block_size - 1, so block_size = K + 1.
+        block_size = self.speculative_config.num_speculative_tokens + 1
```
eplatero97, line 1634: isn't `block_size = self.max_decode_tokens`?

vllm_qaic/spec_decode/dflash_draft_model.py
```diff
@@ -0,0 +1,439 @@
...
+            dtype=np.float32,
+        )
+        self._dlm_batch_index = np.full((self.decode_bsz, 1), -1, dtype=np.int64)
+
+        self._dlm_logits_buf: np.ndarray | None = None
+        self._req_state: dict[str, _ReqState] = {}
+        self._prefill_pending: list[dict] = []
+
+    def load_model(self) -> None:
+        assert self.tlm_prefill_seq_len > 0 and (
```
eplatero97, line 106: I feel like we should move this assertion to `_derive_dflash_config` function so that if this assertion is raised, we don't lose time compiling TLM

vllm_qaic/model_loader/qaic.py
```diff
@@ -156,6 +157,10 @@ def __init__(
         if vllm_config.speculative_config:
             self.num_spec_tokens = vllm_config.speculative_config.num_speculative_tokens
             self.max_decode_tokens += self.num_spec_tokens
+            # DFlash emits block_size logits/step (not 1+K). Public
+            # num_speculative_tokens is block_size - 1, so block_size = K + 1.
+            if vllm_config.speculative_config.method == "dflash":
+                self.max_decode_tokens = self.num_spec_tokens + 1
```
eplatero97, line 163: I left comment in `model_runner.py` that we can use existing logic for this. 

to minimize making two code changes in the future, why don't we create a function that both `qaic.py` and `model_runner.py` both use as the source of truth for computing `max_decode_tokens`?

vllm_qaic/model_loader/qaic.py
```diff
@@ -1304,15 +1355,72 @@ def disagg_dummy_run(self):
...
+        )
+    # TLM captures hidden states after the layer fires, so ids are +1.
+    tlm_hf._dflash_target_layer_ids = [i + 1 for i in target_layer_ids]
+    tlm_hf._dflash_block_size = block_size
+    tlm_hf._dflash_raw_target_layer_ids = list(target_layer_ids)
+    dlm_hf._dflash_block_size = block_size
+    dlm_hf._dflash_mask_token_id = mask_token_id
+    # Cross-checkpoint repos: TLM fetches fc/hidden_norm from DLM; DLM fetches lm_head/embed from TLM.
+    tlm_hf._dflash_dlm_repo = spec_config.draft_model_config.model
+    dlm_hf._dflash_tlm_repo = spec_config.target_model_config.model
```
eplatero97, line 1395: adding qaic-specific compile metadata to shared objects I feel like makes a confusing separation of responsibilities. 

maybe we can move this logic to `_get_qaic_compile_config` or `get_hf_mode` or modify `additional_config` to have a `dflash_cfg` key with all qaic-specific metadata in it?

eplatero97 (changes requested):
In terms of functionality, things looks great. 

New feedback hones in on simplicity, failing fast, and clean separation of responsibilities.

also, let's move dflash unit tests to `tests/test_qaic/spec_decode`

vjanfaza:
> well done, this is looking good!
> 
> I left some final comments.
> 
> can you also please post link to QEff d-flash PR?

https://github.com/quic/efficient-transformers/pull/995

vllm_qaic/worker/model_runner.py
vjanfaza: Yes, fixed!

docs/docs/user_guide/features/speculative_decoding.md
```diff
@@ -118,6 +119,80 @@ docker run --rm -it --network host \
...
+            "prefill_seq_len": block_size,  # DLM prefills block_size at a time
+            "mxfp6_matmul": True,
+            "mxint8_kv_cache": True,
+            "mos": 1,
+        },
+    },
+    speculative_config={
+        "method": "dflash",
+        "model": "z-lab/Qwen3-4B-DFlash-b16",
+        "num_speculative_tokens": block_size,
```
eplatero97: I think this needs to be `block_size-1` now. please check rest of configuration is correct

tests/e2e/test_qaic_dflash.py
```diff
@@ -0,0 +1,162 @@
+# ------------------------------------------------------------------
```
eplatero97: let's move to 'tests/test_qaic/spec_decode/e2e/'

vllm_qaic/worker/model_runner.py
eplatero97: nit

eplatero97 (changes requested):
looks good. please move appropriate unit tests and verify documentation is correct. 

once done, let's assert all dflash unit-tests are passing + 
```
pytest -s tests/e2e/test_qaic_generate.py            # baseline decode + prefill chunking (shared qaic.py/model_runner.py)
pytest -s tests/e2e/test_qaic_output_consistency.py  # regression guard on shared decode output
```

tests/e2e/spec_decode/test_qaic_dflash.py
```diff
@@ -0,0 +1,162 @@
+# ------------------------------------------------------------------
```
eplatero97, line 1: let's move this to `tests/e2e/spec_decode/test_qaic_dflash.py` to follow convention

tests/test_qaic/spec_decode/test_dflash_kv_advance.py
```diff
@@ -0,0 +1,403 @@
...
+Reviewer comment (PR #71): when one request's sequence exceeds
+effective_drafter_max_model_len, the step-wide input_fits_in_drafter gate goes
+False and full draft-proposing is skipped for *every* request that step, not
+just the offending one. Before the fix, QaicDFlashProposer.propose() was only
+ever called when that gate was True, so every other in-decode-phase request's
+DLM KV cache (position_counter) silently fell behind the TLM. This test drives
+QaicDFlashProposer.propose() directly (no QAIC device / QPC needed — the DLM
+forward is replaced by a fake session) to check that a commit=False call still
+advances position_counter for active requests while not offering discarded
+candidates.
```
eplatero97: let's remove this

tests/unit/spec_decode/test_dflash_kv_advance.py
```diff
@@ -0,0 +1,403 @@
+# ------------------------------------------------------------------
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
eplatero97, line 2: let's move this to `tests/unit/spec_decode/test_dflash_kv_advance.py` to follow current convention

tests/test_qaic/spec_decode/test_dflash_kv_advance.py
vjanfaza: Done!

tests/e2e/spec_decode/test_qaic_dflash.py
vjanfaza: Done!

docs/docs/user_guide/features/speculative_decoding.md
vjanfaza: Done!

vjanfaza:
> looks good. please move appropriate unit tests and verify documentation is correct.
> 
> once done, let's assert all dflash unit-tests are passing +
> 
> ```
> pytest -s tests/e2e/test_qaic_generate.py            # baseline decode + prefill chunking (shared qaic.py/model_runner.py)
> pytest -s tests/e2e/test_qaic_output_consistency.py  # regression guard on shared decode output
> ```

Checked all!

eplatero97 (approved):
LGTM

animeshk08 (approved):
LGTM!

PR #74, opened 2026-08-15
Remove long_prefill_token_threshold

quic-sanising:


vllm_qaic/platform_base.py
```diff
@@ -275,9 +275,15 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
         ):
             # QAIC pooling QPCs are compiled with seq_len == max_model_len.
-            # Set both explicitly here (AOT and eager) so prefill_seq_len is correct.
+            # Disable chunked prefill and pin prefill_seq_len to max_model_len via
+            # override_qaic_config (the single source of truth for prefill chunking).
             scheduler_config.enable_chunked_prefill = False
-            scheduler_config.long_prefill_token_threshold = model_config.max_model_len
+            if additional_config is None:
+                additional_config = vllm_config.additional_config = {}
+            additional_config.setdefault("override_qaic_config", {})
```
ankitqti, line 283: _  = additional_config.get("override_qaic_config", {})

vllm_qaic/platform_base.py
```diff
@@ -344,6 +346,13 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
                 additional_config["override_qaic_config"].update(
                     {"prefill_seq_len": __prefill_seq_len}
                 )
+                # Drive the vLLM scheduler prefill chunk from prefill_seq_len so
```
ankitqti, line 349: Remove this

vllm_qaic/platform_base.py
quic-sanising: `setdefault` is the correct behaviour. When key exists, the dictionary remains unchanged, and the existing value is returned. When key is missing, the key is added with the provided default.

vllm_qaic/platform_base.py
ankitqti: any updates on this ?

quic-sanising: Here is what is my understanding:

`prefill_seq_len` and `long_prefill_token_threshold` are genuinely different inputs, but for output accuracy, this line is needed.

</br>

**prefill_seq_len (QEfficient concept):**
- The physical size of the prefill input buffers defined in the QPC. For `vision_embeds`, it is `[vision_batch_size, prefill_seq_len, vision_size]` (e.g.: `[1, 512, 5120]`).
- How many tokens can this compiled binary physically accept in one forward pass? Fixed at compile time; cannot change at runtime.

**long_prefill_token_threshold (vLLM scheduler concept):**
- A scheduler policy: when a prompt is longer than N tokens, chunk it and feed at most N prompt tokens per step.
- The scheduler is the component that decides `num_scheduled_tokens` per request; the QAIC layer only receives whatever the scheduler hands it. If the scheduler decides to process 1271 tokens now, it feeds 1271 embedding tokens into the 512-slot buffer, and `_pad_or_crop` silently truncates tokens 512–1270, leading to junk output.

</br>

The number of embedding tokens fed per step is governed by `long_prefill_token_threshold` and must not exceed the buffer that holds them (`prefill_seq_len`). Setting `long_prefill_token_threshold = prefill_seq_len` makes the software chunk size match the hardware buffer, so each step feeds ≤512 rows → fits → no crop. This is the fix.

  ```python  
  # vllm/v1/core/sched/scheduler.py
  num_new_tokens = request.num_tokens - num_computed_tokens        # e.g. 1271
  if 0 < long_prefill_token_threshold < num_new_tokens:            # 0 < 512 < 1271
      num_new_tokens = long_prefill_token_threshold                # clamp to 512
  num_new_tokens = min(num_new_tokens, token_budget)
  ```

</br>

Previously this value arrived via the `--decode-long-prefill-token-threshold` CLI flag; vLLM's parser wrote it into `scheduler_config.long_prefill_token_threshold`, and our code read it to derive `prefill_seq_len`. This PR removes that user-facing flag, so nothing populates the field anymore. Hence, we now set it internally from `prefill_seq_len`. 
If we don't do so, the default value for `long_prefill_token_threshold` is set, which is 0. When it is 0, `0 < threshold` condition is `false`, so the scheduler never chunks and feeds the full 1271 tokens, which crops against the 512 QPC buffer. So, if we don't set it, it stays 0 and cropping is guaranteed.


</br>

**Can we fix cropping without setting `long_prefill_token_threshold`?** 
In `_gather_mm_embeddings()`, we have this line `encoder_output[start_idx:end_idx]` 
where `end_idx = min(num_computed + num_scheduled_tokens, num_encoder_tokens)`. 

So, the number of gathered embedding tokens is handled by `num_scheduled_tokens`, which is what the scheduler decides per step. 

So, answer is: not really, the crop is fundamentally a scheduler decision (`num_scheduled_tokens`).

ankitqti: Did you tried the same thing using just --max_num_batched_tokens ?

quic-sanising: Yes. It results in same cropping error. 

`max_num_batched_tokens` is a batch-wide budget, whereas we need per-request budget. So, if we set `--max_num_batched_tokens` equal to sequence length (say, 512), the entire batch will get capped to 512 tokens, and not just one request. So, we lose decode batching, we can't run 8 seqs in parallel. This will result in performance degradation. 

And if we set it to batch_size * sequence length, say, 8 * 512 = 4096, then, one request can be scheduled up to the full 4096 tokens. But the `vision_embeds` buffers expects it to be of size 512, therefore, cropping occurs.

ankitqti:
Rebase the PR and run precommit

vllm_qaic/platform_base.py
```diff
@@ -344,6 +346,13 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+                # each prefill step feeds at most prefill_seq_len tokens. This
+                # matches the QPC vision_embeds binding (sized to prefill_seq_len)
+                # and prevents the mm-embedding gather from overflowing/cropping.
+                # (Replaces the removed --*-long-prefill-token-threshold flag.)
+                if not isinstance(__prefill_seq_len, (list, tuple)):
```
animeshk08, line 354: We should not be setting long_prefill_token_threshold at all in the platform file, rather any chunking logic should be generalized in qaic.py forward pass.

animeshk08 (changes requested):
@quic-sanising

PR #75, opened 2026-08-17
[Spec Decode] Enable speculative decoding with disaggregated serving

eplatero97:
# Purpose
This change enables disaggregated serving with speculative decoding methods `ngram`, `suffix`, and `draft_model`.

## MT-Bench performance

The performance results use `meta-llama/Llama-3.1-8B-Instruct` on the 80-prompt `philschmid/mt-bench` workload. The vanilla and disaggregated runs use the same decode Ultra, but disaggregated serving adds a dedicated prefill Ultra. 

The table reports request-throughput change:

| Speculation | `mns=1` | `mns=2` | `mns=4` |
|---|---:|---:|---:|
| None | **+6.76%** | **+7.46%** | **+10.25%** |
| Suffix, K=4 | **+0.15%** | **+6.10%** | **−0.06%** |
| Draft model, K=4 | **+1.78%** | **+5.42%** | **+8.38%** |

> NOTE: draft model was `meta-llama/Llama-3.2-1B`.


ankitqti:
Move this PR to main

ankitqti:
Move this PR to main

tests/e2e/disaggregated_serving/conftest.py
```diff
@@ -135,8 +165,9 @@ def _disagg_test_config(request, pytestconfig, host, port, device_pool_ids):
         "decode_devices": decode_devices,
         "prefill_max_num_seqs": kwargs["prefill_max_num_seqs"],
         "client_request_timeout": pytestconfig.getoption("client_request_timeout"),
-        "disaggregated_startup_timeout": pytestconfig.getoption(
-            "disaggregated_startup_timeout"
+        "disaggregated_startup_timeout": kwargs.get(
```
animeshk08, line 168: Why are there two .get on the same key "disaggregated_startup_timeout"?

tests/test_qaic/disaggregated_serving/test_disagg_spd_unit.py
```diff
@@ -0,0 +1,553 @@
...
+            speculative_config=speculative_config,
+            kv_transfer_config=kv_transfer_config,
+            scheduler_config=SimpleNamespace(max_num_seqs=2, async_scheduling=False),
+            additional_config={},
+        )
+
+    return make
+
+
+def _patch_runner_init():
```
animeshk08, line 46: Why do we need to patch the entire init here? Can't we mock the init and relevant properities instead of patch? I think that will be safe as release progresses.

vllm_qaic/model_loader/qaic.py
```diff
@@ -182,19 +193,26 @@ def __init__(
...
-            _d: dict = {
-                "input_ids": np.full((self.decode_bsz, _mdt), -1, dtype=np.int64),
-                "position_ids": np.full((self.decode_bsz, _mdt), -1, dtype=np.int64),
-                "batch_index": np.full((self.decode_bsz, 1), -1, dtype=np.int64),
-            }
-            if self.lora_mode:
-                _d["lora_ids"] = np.full((self.decode_bsz, 1), -1, dtype=np.int64)
-            self.decode_batch_inputs_by_k[_k] = _d
+            self.decode_logits_by_k[_k] = dict(
+                logits=np.random.randn(self.decode_bsz, _mdt, self.vocab_size).astype(
```
animeshk08: Did we look at which is faster np.random.randn or np.empty?

vllm_qaic/model_loader/qaic.py
```diff
@@ -1556,11 +1542,20 @@ def load_qaic_model(
...
-        # This will create error in parent process if exited,
-        # need better solution in future
-        raise QaicCompilationComplete()
+        if raise_on_compile_complete:
+            # Hack for Model-IP execution flow
+            # TODO: remove this in future
+            # This will create error in parent process if exited,
+            # need better solution in future
+            raise QaicCompilationComplete()
+        # Caller (target load path with a draft model still to compile)
```
animeshk08: Sorry, I did not understand the need for this.. Can you please leave a reply here with more details?

vllm_qaic/worker/model_runner.py
```diff
@@ -408,6 +407,12 @@ class QaicModelRunnerAoT(GPUModelRunner):
         | QaicDraftModelProposer
         | None
     )
+    # Declared here because GPUModelRunner.__init__ sets this dynamically
+    # (from SchedulerConfig.async_scheduling), so mypy cannot infer its type
```
animeshk08: # Declared here because GPUModelRunner.__init__ sets this dynamically
    # (from SchedulerConfig.async_scheduling), so mypy cannot infer its type

> Why is no mypy exception thrown in `GPUModelRunner.__init__ ` then?

animeshk08 (changes requested):
LGTM for most part! As general NIT feedback can we try to keep the inline comments a bit more concise.

Great work Erick!

tests/e2e/disaggregated_serving/conftest.py
eplatero97: the `.get` is actually from two sources:
1. `kwargs.get(...)`
2. `pytestconfig.getoption(...)`

former allows custom values to be passed by user but if empty, we default to `pytestconfig` as was done previously

tests/test_qaic/disaggregated_serving/test_disagg_spd_unit.py
eplatero97: this is a good question and definitely involves some tradeoffs. 

in this unit file we patch only `GPUModelRunner.__init__` (not `QaicModelRunnerAoT.__init__`) so we can still exercise QAIC constructor logic like disagg role handling and decode-K selection.

we bypass the parent initializer because it pulls in model/hardware-dependent setup (e.g., draft proposer, model_config) that is outside this test’s scope and can make unit tests slower/flakier. 

mocking individual parent attributes is possible, but it tends to couple tests more tightly to upstream parent internals, which can become less stable as those internals evolve.

however, let me know what you think on this, I can see it going either way

vllm_qaic/model_loader/qaic.py
eplatero97: likely `np.empty` is faster and also more appropriate given that the actual contents is not significant during initialization. as such, I changed all initializations to `np.empty`

vllm_qaic/model_loader/qaic.py
eplatero97: yep, this is not super-intuitive to read at face-value. 

when using draft model with QPC, if we raise an error right after target model is loaded, draft model will never compile. as such, we do a safe return of target model without loading it so that draft can run it's own `load_qaic_model`

vllm_qaic/worker/model_runner.py
eplatero97: this is a good point. 

`GPUModelRunner.__init__` does set `self.use_async_scheduling` at runtime, but in this QAIC subclass mypy cannot reliably infer that attribute’s type from the dynamic parent initialization path (`super().__init__(...)`).

so we explicitly tell mypy that `use_async_scheduling: bool`

eplatero97 (commented):
good feedback. condensed comments and initialize buffers with empty now. let me know what you think!

PR #76, opened 2026-08-17
 [CI] Add CI support for Eager Mode

loayrash:
## Summary

  Add CI coverage for QAIC PyTorch eager mode, with each E2E job scheduled on its assigned hardware resources.

  - Add a dedicated PYT fasttest entry point and pipeline mode selection.
  - Make shared E2E configuration mode-aware so eager tests do not inherit AoT-only settings; mark unsupported multimodal coverage AoT-only.
  - Scope `QAIC_VISIBLE_DEVICES` to each scheduled job before it starts, preventing eager workers from binding to the wrong physical QID.
  - Give parallel server jobs distinct default ports to avoid contention on `8080`.


ci_scripts/ci_pipeline_run_plugin.sh
```diff
@@ -114,15 +132,17 @@ function execute_tests() {
     mode="$1"
     cd "${VLLM_DIR}"/ci_scripts
     # shellcheck disable=SC1090
-    source ./ci_fasttest_qaic_"${mode}".sh --host "${HOST}" --port "${PORT}"
+    if [ "$mode" = "pyt" ]; then
+        source ./ci_fasttest_qaic_"${mode}".sh --host "${HOST}"
```
animeshk08: Why are the commands different here.. considering the set of tests are similar?

If this is because of the port conflict, similar problem will happen at AoT or Eager tests..

ci_scripts/scheduler.py
```diff
@@ -429,6 +449,14 @@ def main() -> int:
             f"(default: {_DEFAULT_TIMEOUT_S})"
         ),
     )
+    parser.add_argument(
```
animeshk08, line 444: Is there any risk of doing this by default? Why are we setting QAIC_VISIBLE_DEVICES  behind an arg.. I think we can safely set QAIC_VISIBLE_DEVICES in both AoT and Eager..

tests/e2e/multimodal/test_multimodal.py
```diff
@@ -475,6 +475,7 @@ def _qwenvl_multi_resolution_test(
         assert o1.outputs[0].text == o2.outputs[0].text
 
 
+@pytest.mark.qaic_aot_mode
```
animeshk08: Can we extend the qaic_aot_mode to accept optional string arg.. something like 
```
@pytest.mark.qaic_aot_mode("<Reason for why this test is AoT only>")
```

animeshk08 (changes requested):
Looks good for the most part! Left some comments.

Please make sure we used pre-commit linting if not done already

Good work!

ankitqti:
Move this PR to Main

PR #77, opened 2026-08-18
Add support for creation of FullAttentionSpec and SlidingwindowSpec 

quic-vargupt:
This PR adds the support for creating FullAttentionSpec for layers with full attention and SlidingWindowSpec for layers with SWA like Gemma-4 family, gptoss models by parsing the hf_config.


ankitqti:
Move this PR to Main


Also add unit test / e2e test

PR #84, opened 2026-08-24
[main][PyT] Add RotaryEmbedding Support

quic-agokhale:


quic-agokhale:
Register QAicApplyRotaryEmb and QAicRotaryEmbedding as OOT CustomOps,
  delegating the rotation to torch.ops.qaic.rotary_embedding.

vllm_qaic/ops/rotary_embedding.py
```diff
@@ -0,0 +1,196 @@
...
+    concatenated back, so the query/key shapes seen by the caller are preserved.
+
+    ``forward_static`` mirrors the base class staticmethod so that any code calling
+    ``RotaryEmbedding.forward_static`` on a patched instance reaches the QAIC path,
+    and ``forward_oot`` routes regular ``forward`` calls on the instance to that
+    same implementation.
+    """
+
+    @staticmethod
+    def forward_static(
```
ankitqti, line 101: why we need to do forward_static, why not we write everything as forward_oot, so that one function diversion  can be avoided

ankitqti (approved):
Looks good

PR #85, opened 2026-08-24
Replace enforce_eager flag check with is_aot in determining inference mode

quic-agokhale:
Instead of relying on the enforce_eager flag, we should be inspecting the QaicPlatform's is_aot static variable to determine inference mode between AoT and PyT modes. This change does that, ensuring that the QAic plugin can be made consistent with the enforce_eager flag's intended use for torch.compile v/s eager mode determination

### DIFF
--- a/vllm_qaic/platform_base.py
+++ b/vllm_qaic/platform_base.py
@@ -48,8 +48,7 @@ DYNAMIC_RESOLUTION_MODELS = [
 class QaicPlatform(Platform):
     _enum = PlatformEnum.OOT
     primary_attn_backend_cls = (
-        "vllm_qaic.attention.backends"
-        ".qaic_attn.QAicTorchAttentionBackend"
+        "vllm_qaic.attention.backends.qaic_attn.QAicTorchAttentionBackend"
     )
     device_name: str = "qaic"
     # Set device type to cpu if it's AOT.
@@ -125,7 +124,7 @@ class QaicPlatform(Platform):
         if not cls.is_aot:
             return torch_qaic.qaic.get_device_info(device_id).num_cores
         else:
-            pass
+            raise NotImplementedError
 
     @classmethod
     @functools.cache
@@ -133,7 +132,7 @@ class QaicPlatform(Platform):
         if not cls.is_aot:
             return torch_qaic.qaic.get_device_info(device_id).per_core_hvx_thread_count
         else:
-            pass
+            raise NotImplementedError
 
     @classmethod
     def check_if_supports_dtype(cls, dtype: torch.dtype):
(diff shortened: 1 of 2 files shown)


ankitqti:
Rebase and run Pre commit

vllm_qaic/platform_base.py
```diff
@@ -250,11 +240,10 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
         if parallel_config.worker_cls == "auto":
             parallel_config.worker_cls = cls.get_worker_cls()
 
-        if parallel_config.world_size > 1:
-            parallel_config.distributed_executor_backend = "uni"
-            if vllm_config.model_config.enforce_eager:
-                # uni backend sets qualnet ip while mp backend sets localhost ip
-                parallel_config.distributed_executor_backend = "mp"
+        parallel_config.distributed_executor_backend = "uni"
```
quic-sanising: Why is this moved outside the condition?

vllm_qaic/platform_base.py
```diff
@@ -396,7 +385,7 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
         # QAIC_FIXME: default uses torch.compile based custom vllm-backend.
         # Turning off all torch.compile modes to fallback to purely eager for now.
         if compilation_config and compilation_config.mode != CompilationMode.NONE:
-            mode = "eager mode" if vllm_config.model_config.enforce_eager else "AOT"
+            mode = "AoT" if cls.is_aot else "PyT"
```
quic-sanising: The mode strings are changed here. Any downstream effects?

vllm_qaic/platform_base.py
```diff
@@ -200,19 +200,9 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
                 vllm_config.additional_config = cfg
 
         if cls.is_aot:
-            if vllm_config.model_config.enforce_eager:
```
quic-sanising: I think we should keep this warning in any case.

vllm_qaic/platform_base.py
```diff
@@ -200,19 +200,9 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
-                )
-                vllm_config.model_config.enforce_eager = False
+            vllm_config.model_config.enforce_eager = False
             device_config.device = torch.device("cpu")
         else:
-            if not vllm_config.model_config.enforce_eager:
-                logger.warning_once(
-                    "setting inference mode to Eager mode since torch_qaic is installed"
-                )
-                vllm_config.model_config.enforce_eager = True
```
quic-sanising: Should we set this explicitly now?

quic-agokhale: Yes. It should be set explicitly set for strict eager mode. This is in anticipation of the other pytorch-based inference flavors like torch compile and graph execution which also rely on torch_qaic being installed but enforce_eager being false in upstream vllm semantics

vllm_qaic/platform_base.py
quic-agokhale: Moved it back in the if world_size>1 check

vllm_qaic/platform_base.py
quic-agokhale: No downstream effects. Just renamed to keep it consistent with PyT and AoT being the inference flavors.

vllm_qaic/platform_base.py
quic-sanising: Should we keep Line 215?

quic-sanising (commented):
LGTM

PR #93, opened 2026-08-26
Run pre-commit on all files

quic-sanising:
Applies all pre-commit hooks to the existing codebase:

  - ruff: auto-format + fix E501 line wraps and F841 unused variables
  - typos: fix 15 misspellings in comments and docs
  - clang-format: reformat 7 csrc files to Google style
  - markdownlint-cli2: fix MD040 bare fences, MD056 table column count, MD003 heading style, MD036 emphasis-as-heading
  - mypy-3.10/3.12: annotate `update_cfg`, `task`, `prefill_seq_len`, `use_async_scheduling`, `QaicCacheDType`; add assert-not-None guards; fix `Callable | None` and `cast(dict, ...)` patterns
  - validate-config: add docstrings and defaults to `QaicCacheConfig.cache_dtype` and `QaicDeviceConfig fields`


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -17,4 +17,4 @@ A clear and concise description of what you want to happen.
 A clear and concise description of any alternative solutions or features you've considered.
 
 **Additional context**
-Add any other context or screenshots about the feature request here.
```
animeshk08, line 20: Curious what are these diffs? Do we have lint formatting in .md files as well?

animeshk08 (approved):
LGTM!

.github/ISSUE_TEMPLATE/feature_request.md
quic-sanising: Yes, we use `markdownlint`: https://github.com/qualcomm/vllm-qaic/blob/main/.pre-commit-config.yaml#L31.

quic-agokhale (approved):

vllm_qaic/patch/patch_config.py
```diff
@@ -6,30 +6,29 @@
...
 from vllm_qaic.logger import init_logger
 
 logger = init_logger(__name__)
 
-_QAIC_CACHE_DTYPES = ("mxint8",)
-QaicCacheDType = Literal[  # type: ignore[valid-type]
-    tuple(dict.fromkeys((*get_args(CacheDType), *_QAIC_CACHE_DTYPES)))
-]
+# CacheDType is a Literal[str] union in vLLM; alias to str so the type stays
+# valid even when vllm types are unavailable to the type checker.
```
quic-agokhale, line 22: Why did we remove the initialization of CacheDType with the tuple and replace it with a type aliasing of str?

quic-sanising: The upstream vLLM `CacheDType` has strings like "auto", "fp8", etc. We extended it with an extra value: our "mxint8". The problem is that mypy can't evaluate this. mypy expects `Literal` to be fixed value and not a dynamically-constructed tuple. 

Since `CacheDType` is itself ultimately `Literal[str, str, ...]`, which is a subtype of `str`, and "mxint8" is also a `str`, using `str` keeps mypy happy.

quic-agokhale: This still seems too relaxed of a condition from a validation point of view. Instead of validating against the list of supported Cache datatypes, it will accept any valid string as an input cache datatype argument.

Replacing it with   

QaicCacheDType: TypeAlias = Literal[CacheDType, "mxint8"]

will be mypy-friendly as well as strict enough for the validation of passed kv cache dtype.

Need to also import TypeAlias from the typing package at the top-level

quic-agokhale (changes requested):

vllm_qaic/patch/patch_config.py
quic-sanising: Done.

quic-agokhale (approved):

PR #94, opened 2026-08-26
Patch Mooncake connector device id for QAIC

quic-vargupt:


vllm_qaic/patch/__init__.py
```diff
@@ -71,6 +71,17 @@
...
+#
+# ** 6. File: patch_mooncake_connector.py **
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#   1. vllm.distributed.kv_transfer.kv_connector.v1.mooncake.
+#      mooncake_connector.MooncakeConnectorWorker.__init__
+#    Why:
+#       QAIC should not query torch.accelerator.current_device_index() while
+#       initializing the Mooncake connector worker.
+#    How:
+#       Temporarily make current_device_index() return 0 during worker init.
```
animeshk08, line 83: Does this have any risk downstream to default to device idx 0? Do we need the actual device id here?

animeshk08 (changes requested):

PR #101, opened 2026-08-28
Paged Attention AOT 

aditjadh:
**Summary**

This PR adds paged (block-based) KV cache support to the QAIC AOT execution path, which enables prefix caching on vLLM v1. Without paged-attention AOT allocated one contiguous KV region per sequence, sized to the full context length. That meant:

Prefix caching was explicitly unsupported and raised NotImplementedError.
Shared prompt prefixes were recomputed on every request.

With paged attention, KV cache is carved into fixed-size blocks managed by vLLM's block manager and addressed on-device through a per-request block table. Requests only occupy the blocks they need, and identical prefixes can be reused across requests.

**Configuration**
Paged Attention can be enabled by setting --enable-prefix-caching=True.
Block count is derived automatically from the model's context length and prefill chunk size,

**Limitations**
Currently, KV block size is tied to prefill-seq-len; it is not independently configurable.

### DIFF
--- a/vllm_qaic/model_loader/qaic.py
+++ b/vllm_qaic/model_loader/qaic.py
@@ -600,6 +600,25 @@ class QaicCausalLM(nn.Module, SupportsLoRA):
                 )
         return
 
+    def _inject_pa_prefill_inputs(
+        self,
+        chunk_inputs: dict,
+        batch_idx: int,
+        block_table: np.ndarray | None,
+        slot_id: np.ndarray | None,
+        req_index: int,
+    ) -> None:
+        if not self.paged_attention:
+            return
+        if block_table is not None and slot_id is not None:
+            chunk_inputs["block_table"] = block_table[
+                req_index : req_index + 1
+            ].reshape(1, self.num_gpu_blocks_per_batch)
+            chunk_inputs["slot_id"] = slot_id[req_index : req_index + 1]
+        else:
+            chunk_inputs["block_table"] = batch_idx
+            chunk_inputs["slot_id"] = 0
+
     def _run_pipeline_prefill(
         self,
         input_ids: np.ndarray,
@@ -713,17 +732,9 @@ class QaicCausalLM(nn.Module, SupportsLoRA):
                     if logits is not None:
                         chunk_inputs["logits"] = logits[index : index + 1]
 
-                if self.paged_attention:
-                    if block_table is not None and slot_id is not None:
-                        req_block_table = block_table[index : index + 1].reshape(
-                            1, self.num_gpu_blocks_per_batch
-                        )
-                        chunk_inputs["block_table"] = req_block_table
-                        req_slot_id = slot_id[index : index + 1]
-                        chunk_inputs["slot_id"] = req_slot_id
-                    else:
-                        chunk_inputs["block_table"] = batch_index
-                        chunk_inputs["slot_id"] = 0
+                self._inject_pa_prefill_inputs(
+                    chunk_inputs, batch_index, block_table, slot_id, index
+                )
 
                 if pending_exec_count == self.session.prefill_num_execObj:
                     if callback:
@@ -816,17 +827,9 @@ class QaicCausalLM(nn.Module, SupportsLoRA):
                 chunk_inputs["lora_ids"] = lora_index
             if mm_kwargs_list and (mm_kwargs := mm_kwargs_list[i]):
                 chunk_inputs.update(mm_kwargs)
-            if self.paged_attention:
-                if block_table is not None and slot_id is not None:
-                    req_block_table = block_table[i : i + 1].reshape(
-                        1, self.num_gpu_blocks_per_batch
-                    )
-                    chunk_inputs["block_table"] = req_block_table
-                    req_slot_id = slot_id[i : i + 1]
-                    chunk_inputs["slot_id"] = req_slot_id
-                else:
-                    chunk_inputs["block_table"] = batch_index
-                    chunk_inputs["slot_id"] = 0
+            self._inject_pa_prefill_inputs(
+                chunk_inputs, batch_indices[i], block_table, slot_id, i
+            )
             # chunk the request
             n_chunks: int = iids.shape[-1] // self.prefill_seq_len
 
--- a/vllm_qaic/model_loader/qaic.py
+++ b/vllm_qaic/model_loader/qaic.py
@@ -938,20 +938,16 @@ class QaicCausalLM(nn.Module, SupportsLoRA):
 
         if self.paged_attention:
             if block_table is not None and slot_id is not None:
-                self.decode_batch_inputs["block_table"][:num_decodes] = block_table[
-                    :num_decodes
-                ]
-                self.decode_batch_inputs["slot_id"][:num_decodes] = slot_id[
-                    :num_decodes
-                ]
+                batch_inputs["block_table"][:num_decodes] = block_table[:num_decodes]
+                batch_inputs["slot_id"][:num_decodes] = slot_id[:num_decodes]
                 if num_decodes < self.decode_bsz:
-                    self.decode_batch_inputs["block_table"][num_decodes:] = -1
-                    self.decode_batch_inputs["slot_id"][num_decodes:] = 0
+                    batch_inputs["block_table"][num_decodes:] = -1
+                    batch_inputs["slot_id"][num_decodes:] = 0
             else:
-                self.decode_batch_inputs["block_table"][:num_decodes] = batch_indices
+                batch_inputs["block_table"][:num_decodes] = batch_indices
                 if num_decodes < self.decode_bsz:
-                    self.decode_batch_inputs["block_table"][num_decodes:] = -1
-                    self.decode_batch_inputs["slot_id"][num_decodes:] = 0
+                    batch_inputs["block_table"][num_decodes:] = -1
+                    batch_inputs["slot_id"][num_decodes:] = 0
 
         # For spec-decode target: include num_logits_to_keep in batch_inputs
         # so the hardware knows how many token positions to compute logits for.
--- a/vllm_qaic/model_loader/qaic.py
+++ b/vllm_qaic/model_loader/qaic.py
@@ -148,13 +148,12 @@ class QaicCausalLM(nn.Module, SupportsLoRA):
         self.ctx_len = model_config.max_model_len
         self.decode_bsz = vllm_config.scheduler_config.max_num_seqs
         self.full_batch_size = vllm_config.scheduler_config.max_num_seqs
-        self.num_gpu_blocks_per_batch = self.ctx_len // vllm_config.cache_config.block_size
+        # block_size may be None here (vLLM v1 populates it after
+        # determine_num_available_blocks); deferred to load_model.
+        self._vllm_config = vllm_config
+        self.num_gpu_blocks_per_batch = None  # set in load_model
         if vllm_config.cache_config.enable_prefix_caching:
-            self.num_gpu_blocks = (
-                vllm_config.cache_config.num_gpu_blocks_override
-                if vllm_config.cache_config.num_gpu_blocks_override
-                else self.ctx_len // vllm_config.cache_config.block_size
-            )
+            self.num_gpu_blocks = None  # set in load_model
         else:
             self.num_gpu_blocks = self.decode_bsz
         self.paged_attention = bool(vllm_config.cache_config.enable_prefix_caching)
@@ -381,6 +380,18 @@ class QaicCausalLM(nn.Module, SupportsLoRA):
         self.disagg_serving_en = kv_transfer_role is not None
         self.disagg_producer_en = kv_transfer_role == "kv_producer"
 
+        # block_size is now populated (set during cache initialisation)
+        cache_config = self._vllm_config.cache_config
+        self.num_gpu_blocks_per_batch = self.ctx_len // cache_config.block_size
+        if self.paged_attention:
+            self.num_gpu_blocks = (
+                cache_config.num_gpu_blocks_override
+                if cache_config.num_gpu_blocks_override
+                else self.ctx_len // cache_config.block_size
+            )
+        else:
+            self.num_gpu_blocks = self.decode_bsz
+
         logger.info("Loading QPC...")
         logger.info(
             "This may take some time, please don't press CTRL-C during this phase..."
@@ -393,10 +404,20 @@ class QaicCausalLM(nn.Module, SupportsLoRA):
             self.stages = 1
         from .qaic_session_np import QAICInferenceSession
 
+        # qaicrt uses virtual device IDs (0-based within QAIC_VISIBLE_DEVICES).
+        # Translate physical device IDs to virtual IDs before creating the session.
+        _visible = os.environ.get(current_platform.device_control_env_var)
+        if _visible is not None:
+            _visible_ids = [int(x) for x in _visible.replace(",", " ").split()]
+            _phys_to_virt = {phys: virt for virt, phys in enumerate(_visible_ids)}
+            session_device_ids = [_phys_to_virt.get(d, d) for d in device_id]
+        else:
+            session_device_ids = device_id
+
         self.session = QAICInferenceSession(
             qpc_path,
             full_batch_size=self.full_batch_size,
-            device_ids=device_id,
+            device_ids=session_device_ids,
             stages=self.stages,
             cluster_id=cluster_id,
             use_async_scheduling=self.use_async_scheduling,
@@ -925,7 +946,9 @@ class QaicCausalLM(nn.Module, SupportsLoRA):
                 self.decode_batch_inputs["block_table"][:num_decodes] = block_table[
                     :num_decodes
                 ]
-                self.decode_batch_inputs["slot_id"][:num_decodes] = slot_id[:num_decodes]
+                self.decode_batch_inputs["slot_id"][:num_decodes] = slot_id[
+                    :num_decodes
+                ]
                 if num_decodes < self.decode_bsz:
                     self.decode_batch_inputs["block_table"][num_decodes:] = -1
                     self.decode_batch_inputs["slot_id"][num_decodes:] = 0
@@ -1968,10 +1991,22 @@ def _get_qaic_compile_config(
         from qaicrt import QStatus
 
         if cfg["device_group"] is not None:
+            # qaicrt uses virtual device IDs (0-based within QAIC_VISIBLE_DEVICES).
+            # Physical device IDs from device_group must be translated to virtual
+            # IDs before calling getResourceInfo.
+            _visible = os.environ.get(current_platform.device_control_env_var)
+            if _visible is not None:
+                _visible_ids = [int(x) for x in _visible.replace(",", " ").split()]
+                _phys_to_virt = {phys: virt for virt, phys in enumerate(_visible_ids)}
+            else:
+                _phys_to_virt = {}
             for id in cfg["device_group"]:
-                _nsp_info = qaic_util().getResourceInfo(id)
+                virt_id = _phys_to_virt.get(id, id)
+                _nsp_info = qaic_util().getResourceInfo(virt_id)
                 if _nsp_info[0] != QStatus.QS_SUCCESS:
-                    raise ValueError(f"device_id {id} is not available !!")
+                    raise ValueError(
+                        f"device_id {id} (virtual {virt_id}) is not available !!"
+                    )
                 _hw_num_cores = min(_hw_num_cores, _nsp_info[1].nspTotal)
         cfg["num_cores"] = _hw_num_cores
         # Applicable for draft-target spd scheme
@@ -2140,15 +2175,18 @@ def _get_qaic_compile_config(
     # off between producer and consumer stay aligned. Written here, before
     # get_kv_cache_spec() is queried, so the updated value is picked up when
     # the KV cache is sized.
-    if vllm_config.kv_transfer_config:
-        vllm_config.cache_config.block_size = cfg["kv_block_size"]
-    else:
-        vllm_config.cache_config.block_size = cfg["prefill_seq_len"]
 
-    # Add num_kv_blocks through qaic_config
-    num_kv_blocks = vllm_config.model_config.max_model_len // vllm_config.cache_config.block_size
-    logger.info("Num KV Blocks: %s", num_kv_blocks)
     if vllm_config.cache_config.enable_prefix_caching:
+        # Add num_kv_blocks through qaic_config
+        if vllm_config.kv_transfer_config:
+            num_kv_blocks = (
+                vllm_config.model_config.max_model_len // cfg["kv_block_size"]
+            )
+        else:
+            num_kv_blocks = (
+                vllm_config.model_config.max_model_len // cfg["prefill_seq_len"]
+            )
+        logger.info("Num KV Blocks: %s", num_kv_blocks)
         qaic_config["num_kv_blocks"] = num_kv_blocks
         qaic_config["blocking_mode"] = "kv_paged"
         qaic_config["enable_blocking"] = True
--- a/vllm_qaic/model_loader/qaic.py
+++ b/vllm_qaic/model_loader/qaic.py
@@ -148,12 +148,12 @@ class QaicCausalLM(nn.Module, SupportsLoRA):
         self.ctx_len = model_config.max_model_len
         self.decode_bsz = vllm_config.scheduler_config.max_num_seqs
         self.full_batch_size = vllm_config.scheduler_config.max_num_seqs
-        self.num_gpu_blocks_per_batch = self.ctx_len // self.prefill_seq_len
+        self.num_gpu_blocks_per_batch = self.ctx_len // vllm_config.cache_config.block_size
         if vllm_config.cache_config.enable_prefix_caching:
             self.num_gpu_blocks = (
                 vllm_config.cache_config.num_gpu_blocks_override
                 if vllm_config.cache_config.num_gpu_blocks_override
-                else self.ctx_len // self.prefill_seq_len
+                else self.ctx_len // vllm_config.cache_config.block_size
             )
         else:
             self.num_gpu_blocks = self.decode_bsz
@@ -2134,17 +2134,22 @@ def _get_qaic_compile_config(
                 if (len(cfg.get("comp_ctx_lengths_decode", [])) == 0)
                 else cfg.get("comp_ctx_lengths_decode")
             )
+    # For disaggregated serving, the KV cache block size must match the
+    # prefill chunk size actually compiled into the QPC (post-override), not
+    # just the scheduler's long_prefill_token_threshold, so KV blocks handed
+    # off between producer and consumer stay aligned. Written here, before
+    # get_kv_cache_spec() is queried, so the updated value is picked up when
+    # the KV cache is sized.
+    if vllm_config.kv_transfer_config:
+        vllm_config.cache_config.block_size = cfg["kv_block_size"]
+    else:
+        vllm_config.cache_config.block_size = cfg["prefill_seq_len"]
+
     # Add num_kv_blocks through qaic_config
-    logger.info(
-        "Num KV Blocks: %s",
-        vllm_config.model_config.max_model_len
-        // vllm_config.scheduler_config.long_prefill_token_threshold,
-    )
+    num_kv_blocks = vllm_config.model_config.max_model_len // vllm_config.cache_config.block_size
+    logger.info("Num KV Blocks: %s", num_kv_blocks)
     if vllm_config.cache_config.enable_prefix_caching:
-        qaic_config["num_kv_blocks"] = (
-            vllm_config.model_config.max_model_len
-            // vllm_config.scheduler_config.long_prefill_token_threshold
-        )
+        qaic_config["num_kv_blocks"] = num_kv_blocks
         qaic_config["blocking_mode"] = "kv_paged"
         qaic_config["enable_blocking"] = True
     else:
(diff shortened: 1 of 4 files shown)


ankitqti:
Too many lines, take only 50 lines budget it should be not more than that

vllm_qaic/model_loader/qaic.py
```diff
@@ -148,6 +148,13 @@ def __init__(
         self.ctx_len = model_config.max_model_len
         self.decode_bsz = vllm_config.scheduler_config.max_num_seqs
         self.full_batch_size = vllm_config.scheduler_config.max_num_seqs
+        self._vllm_config = vllm_config
+        self.num_gpu_blocks_per_batch = None
```
ankitqti: Why you need this here?

vllm_qaic/model_loader/qaic.py
```diff
@@ -364,6 +378,18 @@ def load_model(
         self.disagg_serving_en = kv_transfer_role is not None
         self.disagg_producer_en = kv_transfer_role == "kv_producer"
 
+        # block_size is now populated (set during cache initialisation)
+        cache_config = self._vllm_config.cache_config
```
ankitqti: No need for this

vllm_qaic/model_loader/qaic.py
```diff
@@ -376,10 +402,20 @@ def load_model(
             self.stages = 1
         from .qaic_session_np import QAICInferenceSession
 
+        # qaicrt uses virtual device IDs (0-based within QAIC_VISIBLE_DEVICES).
+        # Translate physical device IDs to virtual IDs before creating the session.
+        _visible = os.environ.get(current_platform.device_control_env_var)
+        if _visible is not None:
```
ankitqti: No need for this

vllm_qaic/model_loader/qaic.py
```diff
@@ -393,6 +429,14 @@ def load_model(
             np.dtype(_logits_info[1]) if _logits_info is not None else np.float32
         )
 
+        if "block_table" in self.session.input_names:
```
ankitqti: Fix this get size of the block from the qpc

vllm_qaic/model_loader/qaic.py
```diff
@@ -667,6 +713,18 @@ def _run_pipeline_prefill(
                     if logits is not None:
                         chunk_inputs["logits"] = logits[index : index + 1]
 
+                if self.paged_attention:
```
ankitqti: No need for this, all options lands to chunk_input any ways

vllm_qaic/model_loader/qaic.py
```diff
@@ -756,6 +816,17 @@ def _run_prefill(
                 chunk_inputs["lora_ids"] = lora_index
             if mm_kwargs_list and (mm_kwargs := mm_kwargs_list[i]):
                 chunk_inputs.update(mm_kwargs)
+            if self.paged_attention:
```
ankitqti: No need for this, all options lands to chunk_input any ways

ankitqti: No need for this, all options lands to chunk_input any ways

vllm_qaic/model_loader/qaic.py
```diff
@@ -866,6 +939,23 @@ def _run_decode(
             if num_decodes < self.decode_bsz:
                 batch_inputs["lora_ids"][num_decodes:] = -1
 
+        if self.paged_attention:
```
ankitqti: No need for this, all options lands to chunk_input any ways

ankitqti:
Lot of bloated code, just less than 50~75 lines overall you need

vllm_qaic/model_loader/qaic.py
```diff
@@ -376,10 +402,20 @@ def load_model(
             self.stages = 1
         from .qaic_session_np import QAICInferenceSession
 
+        # qaicrt uses virtual device IDs (0-based within QAIC_VISIBLE_DEVICES).
+        # Translate physical device IDs to virtual IDs before creating the session.
+        _visible = os.environ.get(current_platform.device_control_env_var)
```
quic-vargupt, line 401: Why are we adding this new env?

vllm_qaic/model_loader/qaic.py
```diff
@@ -148,6 +148,13 @@ def __init__(
         self.ctx_len = model_config.max_model_len
         self.decode_bsz = vllm_config.scheduler_config.max_num_seqs
         self.full_batch_size = vllm_config.scheduler_config.max_num_seqs
+        self._vllm_config = vllm_config
```
quic-vargupt: no need for this line

vllm_qaic/model_loader/qaic.py
quic-vargupt: we can directly take cache_config from vllm_config

vllm_qaic/model_loader/qaic.py
```diff
@@ -364,6 +378,18 @@ def load_model(
...
         self.disagg_producer_en = kv_transfer_role == "kv_producer"
 
+        # block_size is now populated (set during cache initialisation)
+        cache_config = self._vllm_config.cache_config
+        self.num_gpu_blocks_per_batch = self.ctx_len // cache_config.block_size
+        if self.paged_attention:
+            self.num_gpu_blocks = (
+                cache_config.num_gpu_blocks_override
+                if cache_config.num_gpu_blocks_override
+                else self.ctx_len // cache_config.block_size
```
quic-vargupt: Let's make it (ctx_len + block_size - 1) // block_size, for example for ctx_len 1400 and block_size 150, we need 10 blocks but this calculation will return 9

vllm_qaic/model_loader/qaic.py
```diff
@@ -866,6 +939,23 @@ def _run_decode(
...
                 batch_inputs["lora_ids"][num_decodes:] = -1
 
+        if self.paged_attention:
+            if block_table is not None and slot_id is not None:
+                self.decode_batch_inputs["block_table"][:num_decodes] = block_table[
+                    :num_decodes
+                ]
+                self.decode_batch_inputs["slot_id"][:num_decodes] = slot_id[
+                    :num_decodes
+                ]
```
quic-vargupt: Can you Please check line 942-949 again?

vllm_qaic/model_loader/qaic.py
```diff
@@ -2067,6 +2167,32 @@ def _get_qaic_compile_config(
...
+        if vllm_config.kv_transfer_config:
+            num_kv_blocks = (
+                vllm_config.model_config.max_model_len // cfg["kv_block_size"]
+            )
+            vllm_config.cache_config.block_size = cfg["kv_block_size"]
+        else:
+            num_kv_blocks = (
+                vllm_config.model_config.max_model_len // cfg["prefill_seq_len"]
+            )
+            vllm_config.cache_config.block_size = cfg["prefill_seq_len"]
```
quic-vargupt, line 2180: Why are we having 2 different parameters kv_block_size and prefill_seq_len for disaggregated and non-disaggregated serving? Plus we also need to remove kv_block_size from qeff compile kwargs if we're going to use it.

vllm_qaic/worker/model_runner.py
```diff
@@ -1547,6 +1591,15 @@ def load_model(self, *args, **kwargs) -> None:
...
                 )
+        # if self.model.paged_attention:
+        #     decode_block_table = np.arange(
+        #         self.model.decode_bsz * self.model.num_gpu_blocks_per_batch,
+        #         dtype=np.int64,
+        #     ).reshape(self.model.decode_bsz, self.model.num_gpu_blocks_per_batch)
+        #     decode_slot_ids = np.zeros(self.model.decode_bsz, dtype=np.int64)
+        # else:
+        #     decode_block_table = None
+        #     decode_slot_ids = None
```
quic-vargupt: why are we commenting out this code?

vllm_qaic/worker/worker.py
```diff
@@ -702,11 +726,24 @@ def reload_weights(self) -> None:
...
+                    self.scheduler_config.max_num_seqs
+                    * self.cache_config.num_gpu_blocks_override
+                    + 1
+                )
+            else:
+                num_gpu_blocks = (
+                    self.scheduler_config.max_num_seqs
+                    * (
+                        self.vllm_config.model_config.max_model_len
+                        // self.cache_config.block_size
```
quic-vargupt: Let's take ceil value of this division

vllm_qaic/platform_base.py
```diff
@@ -539,6 +548,16 @@ def _configure_multimodal_model(
...
+            cache_config = vllm_config.cache_config
+            if cache_config and cache_config.enable_prefix_caching:
+                if vllm_config.kv_transfer_config:
+                    cache_config.block_size = vllm_config.additional_config.get(
+                        "override_qaic_config", {}
+                    ).get("kv_block_size")
+                else:
+                    cache_config.block_size = vllm_config.additional_config.get(
+                        "override_qaic_config", {}
+                    ).get("prefill_seq_len")
```
quic-vargupt: This code is repeated everywhere, can't we have a helper function for this?

quic-vargupt (changes requested):

vllm_qaic/model_loader/qaic.py
aditjadh: As discussed, prefill and decode will end up with different blocks sizes during DA

vllm_qaic/model_loader/qaic.py
aditjadh: This was a temp workaround, will revert to original

vllm_qaic/model_loader/qaic.py
```diff
@@ -2067,6 +2164,31 @@ def _get_qaic_compile_config(
...
+    # prefill chunk size actually compiled into the QPC (post-override), not
+    # just the scheduler's long_prefill_token_threshold, so KV blocks handed
+    # off between producer and consumer stay aligned. Written here, before
+    # get_kv_cache_spec() is queried, so the updated value is picked up when
+    # the KV cache is sized.
+
+    if vllm_config.cache_config.enable_prefix_caching:
+        # Add num_kv_blocks through qaic_config
+        if vllm_config.kv_transfer_config:
+            kv_block_size = cfg.pop("kv_block_size")
```
quic-vargupt, line 2173: inside qaic.py, when we are calling qeff_model.compile, the parameter kv_block_size is still going to the compiler in case of disaggregated serving causing both P and D compilation to fail, can you please check once if we're popping it correctly. Thanks

quic-vargupt (changes requested):

vllm_qaic/model_loader/qaic.py
```diff
@@ -148,6 +148,8 @@ def __init__(
         self.ctx_len = model_config.max_model_len
         self.decode_bsz = vllm_config.scheduler_config.max_num_seqs
         self.full_batch_size = vllm_config.scheduler_config.max_num_seqs
+        self.paged_attention = bool(vllm_config.cache_config.enable_prefix_caching)
```
athavale-shivani: self.paged_attention is derived here from cache_config.enable_prefix_caching, but decode/prefill later also gate on "block_table" in self.session.input_names (whether the compiled QPC actually exposes that input). If a stale QPC and this flag ever disagree, could hit a KeyError downstream in _run_decode/_inject_pa_prefill_inputs. Might be worth asserting the two agree right here at load time.

vllm_qaic/model_loader/qaic.py
```diff
@@ -268,7 +276,8 @@ def forward(
                         batch_indices,
                         logits,
                         lora_ids,
-                        callback=callback,
```
athavale-shivani: This call used to pass callback=callback — now that's gone and block_table/slot_id are positional instead. _run_decode's signature still has callback as a param, should probably keep callback=callback here alongside block_table/slot_id.

vllm_qaic/model_loader/qaic.py
```diff
@@ -363,6 +372,17 @@ def load_model(
         self.stages: int = stages if stages is not None else 1
         self.disagg_serving_en = kv_transfer_role is not None
         self.disagg_producer_en = kv_transfer_role == "kv_producer"
+        self.num_gpu_blocks_per_batch = cdiv(
+            self.ctx_len, self._cache_config.block_size
+        )
+        if self.paged_attention:
```
athavale-shivani: self.num_gpu_blocks doesn't appear to be read anywhere after it's set here — could this block be dropped, or is it needed for something?

vllm_qaic/model_loader/qaic.py
```diff
@@ -376,10 +396,20 @@ def load_model(
             self.stages = 1
         from .qaic_session_np import QAICInferenceSession
 
+        # qaicrt uses virtual device IDs (0-based within QAIC_VISIBLE_DEVICES).
```
athavale-shivani: The changes from line 399-412 look unrelated to paged attention — could this be split out into a separate PR? Would make it easier to review independently.

vllm_qaic/model_loader/qaic.py
```diff
@@ -393,6 +423,14 @@ def load_model(
             np.dtype(_logits_info[1]) if _logits_info is not None else np.float32
         )
 
+        if "block_table" in self.session.input_names:
```
athavale-shivani: This gates on "block_table" in self.session.input_names (whether the compiled QPC exposes that input), which is a separate signal from self.paged_attention (set earlier from cache_config.enable_prefix_caching). If a stale QPC and the flag ever disagree, could cause a KeyError downstream in _run_decode. Might be worth asserting the two agree at load time, or unifying into a single source of truth.

vllm_qaic/model_loader/qaic.py
```diff
@@ -556,6 +594,25 @@ def _kv_handoff(
                 )
         return
 
+    def _inject_pa_prefill_inputs(
```
athavale-shivani: When self.paged_attention is true, every call path I can find always passes non-None block_table/slot_id. So this else branch (chunk_inputs["block_table"] = batch_idx, a scalar) looks unreachable — is there a call path that hits it, or can it be dropped?

vllm_qaic/model_loader/qaic.py
```diff
@@ -1899,10 +1982,22 @@ def _get_qaic_compile_config(
         from qaicrt import QStatus
 
         if cfg["device_group"] is not None:
+            # qaicrt uses virtual device IDs (0-based within QAIC_VISIBLE_DEVICES).
```
athavale-shivani: Line 1985-2000 look unrelated to paged attention — as suggested above for another device related code block, could this be split out into a separate PR? Would make it easier to review independently.

vllm_qaic/model_loader/qaic.py
```diff
@@ -866,6 +936,19 @@ def _run_decode(
...
                 batch_inputs["lora_ids"][num_decodes:] = -1
 
+        if self.paged_attention:
+            if block_table is not None and slot_id is not None:
+                batch_inputs["block_table"][:num_decodes] = block_table[:num_decodes]
+                batch_inputs["slot_id"][:num_decodes] = slot_id[:num_decodes]
+                if num_decodes < self.decode_bsz:
+                    batch_inputs["block_table"][num_decodes:] = -1
+                    batch_inputs["slot_id"][num_decodes:] = 0
+            else:
```
athavale-shivani: Same question as on _inject_pa_prefill_inputs — when self.paged_attention is true, is this else branch ever reachable (e.g. via qaic_draft_model.py's calls, which don't pass block_table/slot_id)? 

If so, a couple of things stand out: batch_inputs["block_table"][:num_decodes] =batch_indices assigns request batch-positions into the block-table slots. Since block_table[rows, block_index] is used directly as the physical block ID on the QEfficient side (cache_utils.py:538), this could scatter into whatever block happens to share that number, not one actually allocated to the request. Separately, this assignment only broadcasts correctly if num_decodes happens to equal num_gpu_blocks_per_batch (or is 1) — otherwise it'd raise a shape error. And when num_decodes == self.decode_bsz, this branch sets block_table but never initializes slot_id — is that intentional?

vllm_qaic/model_loader/qaic.py
```diff
@@ -2067,6 +2160,31 @@ def _get_qaic_compile_config(
...
+    # prefill chunk size actually compiled into the QPC (post-override), not
+    # just the scheduler's long_prefill_token_threshold, so KV blocks handed
+    # off between producer and consumer stay aligned. Written here, before
+    # get_kv_cache_spec() is queried, so the updated value is picked up when
+    # the KV cache is sized.
+
+    if vllm_config.cache_config.enable_prefix_caching:
+        # Add num_kv_blocks through qaic_config
+        if vllm_config.kv_transfer_config:
+            kv_block_size = cfg.pop("kv_block_size")
```
athavale-shivani: _pa_block_size() uses .get() (silently returns None if the key's missing), while this uses cfg.pop("kv_block_size") (raises a bare KeyError). Worth reconciling so a missing kv_block_size fails the same way in both places?

vllm_qaic/model_loader/qaic.py
```diff
@@ -2067,6 +2160,31 @@ def _get_qaic_compile_config(
...
+            num_kv_blocks = cdiv(vllm_config.model_config.max_model_len, kv_block_size)
+            vllm_config.cache_config.block_size = kv_block_size
```
athavale-shivani: cache_config.block_size = kv_block_size here looks redundant with platform_base.py::_pa_block_size(), which already sets cache_config.block_size from the same override_qaic_config["kv_block_size"] earlier in check_and_update_config. Is this second assignment intentional (e.g. a safety re-assert), or can it be dropped since _pa_block_size() already covers it?

vllm_qaic/model_loader/qaic.py
athavale-shivani: The comment above this block says the KV cache block size "must match the prefill chunk size... so KV blocks handed off between producer and consumer stay aligned," but here you mention that "prefill and decode will end up with different block sizes during DA." Those seem to be in tension — if producer and consumer really do end up with different block sizes, what does "stay aligned" mean here? Is it referring to something other than producer-size == consumer-size (e.g. token-boundary alignment), and if sizes can differ, how does the KV handoff in qaic_connector.py tolerate that mismatch?

vllm_qaic/model_loader/qaic.py
```diff
@@ -2067,6 +2160,31 @@ def _get_qaic_compile_config(
...
+        else:
+            num_kv_blocks = cdiv(
+                vllm_config.model_config.max_model_len, cfg["prefill_seq_len"]
+            )
+            vllm_config.cache_config.block_size = cfg["prefill_seq_len"]
+        logger.info("Num KV Blocks: %s", num_kv_blocks)
+        qaic_config["num_kv_blocks"] = num_kv_blocks
+        qaic_config["blocking_mode"] = "kv_paged"
```
athavale-shivani: If a caller sets blocking_mode in qaic_config directly, would this line silently overwrite it to "kv_paged" whenever prefix caching is on?

vllm_qaic/model_loader/qaic.py
```diff
@@ -2067,6 +2160,31 @@ def _get_qaic_compile_config(
...
+        qaic_config["enable_blocking"] = True
```
athavale-shivani: qaic_config["enable_blocking"] is this required downstream?

vllm_qaic/worker/model_runner.py
```diff
@@ -815,6 +815,32 @@ def _prepare_qaic_inputs(
...
+            # vLLM v0.23's expects (num_reqs, query_start_loc, positions) signature.
+            query_start_loc_np = np.concatenate(([0], cu_num_tokens[:num_reqs])).astype(
+                np.int32
+            )
+            self.input_batch.block_table.compute_slot_mapping(
+                num_reqs,
+                torch.from_numpy(query_start_loc_np),
+                torch.from_numpy(positions_np),
+            )
+            self.slot_mapping = (
```
athavale-shivani: self.slot_mapping is assigned here but doesn't appear to be read anywhere else in the repo — qaic_attn.py/qaic_attn_fullblocks.py read slot_mapping off common_attn_metadata, not off this instance attribute. Is self.slot_mapping needed for something not visible from this file, or can this assignment be dropped (while keeping the compute_slot_mapping(...) call itself, since that populates input_batch's own slot_mapping that attention metadata does depend on)?

vllm_qaic/model_loader/qaic.py
athavale-shivani: Pass qaic_config to compile.

vllm_qaic/worker/worker.py
```diff
@@ -630,19 +631,31 @@ def execute_dummy_batch(self) -> None:
 
 
 class QaicWorkerAoT(QaicWorker):
+    def _compute_num_gpu_blocks(self) -> int:
+        if self.cache_config.enable_prefix_caching:
+            blocks_per_seq = self.cache_config.num_gpu_blocks_override or cdiv(
```
athavale-shivani: There appear to be three separate places that independently recompute what's meant to be the same "blocks needed per sequence" quantity via cdiv(ctx_len_or_max_model_len, block_size):
- qaic.py's _get_qaic_compile_config() (num_kv_blocks, passed to the compiler)
- qaic.py's QaicCausalLM.load_model() (self.num_gpu_blocks_per_batch / self.num_gpu_blocks, sizing the host-side block_table/slot_id buffers)
- worker.py's _compute_num_gpu_blocks() (blocks_per_seq, feeding cache_config.num_gpu_blocks for vLLM's scheduler)

Only the last one consults cache_config.num_gpu_blocks_override; the other two ignore it. If someone sets num_gpu_blocks_override, vLLM's scheduler would end up believing each sequence has a different number of blocks available than what's actually compiled into the QPC and what the host block_table buffers are sized for. Since QaicCausalLM.load_model() already computes an override-aware self.num_gpu_blocks (qaic.py:378-381) in the same function where num_gpu_blocks_per_batch is computed, would it make sense for num_gpu_blocks_per_batch to just reuse that value instead of recomputing independently? And should _get_qaic_compile_config()'s num_kv_blocks also respect the override, so the compiled cache size, the host buffers, and the scheduler's view all stay in sync?

vllm_qaic/model_loader/qaic.py
```diff
@@ -2067,6 +2160,31 @@ def _get_qaic_compile_config(
...
+    # just the scheduler's long_prefill_token_threshold, so KV blocks handed
+    # off between producer and consumer stay aligned. Written here, before
+    # get_kv_cache_spec() is queried, so the updated value is picked up when
+    # the KV cache is sized.
+
+    if vllm_config.cache_config.enable_prefix_caching:
+        # Add num_kv_blocks through qaic_config
+        if vllm_config.kv_transfer_config:
+            kv_block_size = cfg.pop("kv_block_size")
+            num_kv_blocks = cdiv(vllm_config.model_config.max_model_len, kv_block_size)
```
athavale-shivani: These two computations, plus worker.py's _compute_num_gpu_blocks(), all independently derive what's meant to be the same "blocks needed per sequence" value via cdiv(ctx_len_or_max_model_len, block_size). Only _compute_num_gpu_blocks() consults cache_config.num_gpu_blocks_override; num_kv_blocks (baked into the compiled QPC) and num_gpu_blocks_per_batch (sizing the host block_table/slot_id buffers) don't. Since load_model() already computes an override-aware self.num_gpu_blocks (a few lines above num_gpu_blocks_per_batch, in the same function), could num_gpu_blocks_per_batch just reuse that value instead of recomputing independently? And should num_kv_blocks also respect the override, so the compiled cache size and the host buffers can't silently drift out of sync with whatever value the override forces?

vllm_qaic/worker/worker.py
```diff
@@ -630,19 +631,31 @@ def execute_dummy_batch(self) -> None:
...
+            return self.scheduler_config.max_num_seqs * blocks_per_seq + 1
+        return self.scheduler_config.max_num_seqs + 1
+
     def initialize_cache(self, num_gpu_blocks: int, num_cpu_blocks: int) -> None:
         self.cache_config.num_cpu_blocks = num_cpu_blocks
         # disable sliding window
         self.cache_config.sliding_window = None
 
         assert num_cpu_blocks == 0
+        if self.model_config.enforce_eager:
```
athavale-shivani: QaicPlatform.check_and_update_config (platform_base.py:200-207) unconditionally forces enforce_eager = False whenever is_aot is true, and QaicWorkerAoT is only ever selected as the worker class when is_aot is true. Nothing else in the repo sets enforce_eager back to True for the AoT path. Doesn't that make this if self.model_config.enforce_eager: branch on QaicWorkerAoT unreachable? If there's a way to get enforce_eager=True into an AoT worker that I'm missing, could you point to it — otherwise should this branch be dropped?

vllm_qaic/worker/worker.py
```diff
@@ -630,19 +631,31 @@ def execute_dummy_batch(self) -> None:
...
 class QaicWorkerAoT(QaicWorker):
+    def _compute_num_gpu_blocks(self) -> int:
+        if self.cache_config.enable_prefix_caching:
+            blocks_per_seq = self.cache_config.num_gpu_blocks_override or cdiv(
+                self.model_config.max_model_len, self.cache_config.block_size
+            )
+            return self.scheduler_config.max_num_seqs * blocks_per_seq + 1
+        return self.scheduler_config.max_num_seqs + 1
+
     def initialize_cache(self, num_gpu_blocks: int, num_cpu_blocks: int) -> None:
```
athavale-shivani: Is initialize_cache still reachable in this codebase? Under vLLM v1 the cache-sizing RPC that actually gets dispatched is initialize_from_config (already implemented separately on QaicWorker/QaicWorkerPyt), not initialize_cache — I don't see initialize_cache referenced anywhere in vLLM's engine-core/executor RPC dispatch, nor called from anywhere else in this repo. If it's dead code, the new _compute_num_gpu_blocks() logic and paged-attention branch added here wouldn't actually run. If there's a call path I'm missing, could you point to it?

vllm_qaic/worker/worker.py
```diff
@@ -630,19 +631,31 @@ def execute_dummy_batch(self) -> None:
...
 
         assert num_cpu_blocks == 0
+        if self.model_config.enforce_eager:
+            # adapted from gpu_worker.py
+            self.cache_config.num_gpu_blocks = num_gpu_blocks
+            return
+
         if not self.cache_config.enable_prefix_caching:
             self.cache_config.num_gpu_blocks = num_gpu_blocks
-            # Sanity check: AOT requires exact block count; eager is flexible
```
athavale-shivani: why is this comment dropped?

vllm_qaic/platform_base.py
```diff
@@ -301,16 +301,25 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
                     cache_config.mamba_block_size = (
                         model_config.max_model_len
                     )  # reset to "no-op" default
                     cache_config.mamba_cache_mode = "none"  # reset to disabled
-                    logger.warning_once(
-                        "Prefix caching is not yet supported on v1 Engine. "
-                        "Will automatically disable it."
-                    )
-                cache_config.block_size = model_config.max_model_len  # ctx_len
+                    cache_config.block_size = cls._pa_block_size(vllm_config)
```
athavale-shivani: On line 308, cls._pa_block_size(vllm_config) reads override_qaic_config.get("prefill_seq_len") for the non-kv_transfer_config path. But the default value of  prefill_seq_len (when the caller hasn't supplied one explicitly) is only computed and written into override_qaic_config later in this same method, at lines ~334–354 (the __prefill_seq_len block under if cls.is_aot:). So for enable_prefix_caching=True + no kv_transfer_config + no explicit prefill_seq_len passed in additional_config, line 308 runs before that default exists, and _pa_block_size returns None — setting cache_config.block_size = None. This only happens to work today because probably every existing example/test that sets enable_prefix_caching=True also explicitly passes prefill_seq_len in override_qaic_config up front so the value already exists when line 308 runs. Should the default-computation block (lines 334–354) run before this block_size assignment?

vllm_qaic/platform_base.py
```diff
@@ -301,16 +301,25 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+                elif (
+                    model_config.is_multimodal_model and model_config.is_encoder_decoder
+                ):
+                    cache_config.block_size = 100000
```
athavale-shivani: Where does 100000 come from for the multimodal block_size?

vllm_qaic/platform_base.py
```diff
@@ -301,16 +301,25 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+                else:
+                    cache_config.block_size = model_config.max_model_len  # ctx_len
+
+        # disable async scheduling since Qaic does not yet support it
```
athavale-shivani: Why are we disabling async scheduling here?

vllm_qaic/platform_base.py
```diff
@@ -539,6 +548,11 @@ def _configure_multimodal_model(
                 scheduler_config.max_num_seqs
                 * scheduler_config.long_prefill_token_threshold
             )
+            if (
```
athavale-shivani: This recomputes cls._pa_block_size(vllm_config) and reassigns cache_config.block_size, but for is_vision_encoder and cls.is_aot, line 308 (earlier in this same check_and_update_config call) already did this identically — same method, same vllm_config, no changed inputs in between. Is this line needed, or is it a leftover duplicate of the line-308 assignment? (Also inherits the same ordering issue: prefill_seq_len isn't written into override_qaic_config until line 559, right after this call, so if the caller didn't pass it explicitly, both this and line 308 read None.)

vllm_qaic/platform_base.py
```diff
@@ -422,13 +431,6 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
                 "serving, other SPD types such as Turbo is not yet supported "
                 "with Disaggregated serving for QAIC backend"
             )
-            assert not (
```
athavale-shivani: This assert blocking enable_prefix_caching with kv_role != "kv_producer" is removed with no replacement check (and no e2e test currently exercises kv_consumer  + enable_prefix_caching=True). Since kv_both is still independently blocked in qaic_connector.py:353, this concretely unlocks kv_consumer + enable_prefix_caching=True starting now, not just once the NIXL work lands — is that intentional, or should this restriction stay in place until the NIXL PR is in and that combination is validated?

PR #98, opened 2026-08-28
Rebase vLLM to v0.28.0

quic-sanising:
Rebases the QAIC plugin from vLLM v0.23.0 to v0.28.0.

  ### Version bumps

  - `setup.py`, `pyproject.toml`, `scripts/utility.sh`: v0.23.0 → v0.28.0
  - `requirements/vllm_dependency_{aot,pyt}.txt`: aligned with upstream v0.28.0 (transformers, fastapi/starlette, prometheus-fastapi-instrumentator, xgrammar, mistral_common, jsonschema; dropped diskcache/gguf)
  - `requirements/lint.txt`: `pre-commit` → `>=4.5.1`

  ### v0.28.0 API adaptations

  - `patch_parallel_state.py`: add `use_all2all` arg to `QaicGroupCoordinator` (new `GroupCoordinator` param)
  - `worker/model_runner.py`: unpack new `_num_nans_device` from `_bookkeeping_sync`; set `self.pin_memory = False` (upstream moved to module-level `PIN_MEMORY`); patch `torch.cuda.Event` in `_torch_cuda_wrapper`
  - `patch_block_table.py`: handle new `SlotMappingMode` (early-return for `NONE`)
  - `patch_rejection_sampler.py`: pass `synthetic_mode`/`synthetic_conditional_rates`/`use_fp64_gumbel` to `rejection_sample`; drop stale skip-softmax path
  - `patch_mem_utils.py`: capability-guard platform memory hooks + add `MemorySnapshot.__post_init__` override
  - `__init__.py`: apply graph-pickler `Options` shim at import (torch < 2.8 compat)

  ### Transformers 5.15.1 compatibility

  - `qaic_custom_mm_processor.py`: fallback import for `Qwen2VLImageProcessorFast`; widen version gate to
    `>= 5.5.4`

  ### Test / CI

  - `test_qaic_lora.py`: fix swapped `seq_len`/`ctx_len` (violated new `long_prefill_token_threshold ≤ max_model_len` check)
  - `requirements/test.txt`: consolidate test deps (drop linters, add datasets)
  - `ci_pipeline_run_plugin.sh`: install test deps from `requirements/test.txt` (single source of truth)

  ### Validation

  - `examples/qaic.py` runs end-to-end in AOT
  - AOT CI (`ci_fasttest_qaic_aot.sh`) passes

PR #104, opened 2026-08-30
Use V1 model runner for v0.28 PYT warmup

loayrash:
## Summary

Restore the normal vLLM kernel warmup path for QAIC PYT and explicitly select the V1 model runner:

- Set `QaicWorkerPyt.use_v2_model_runner = False`.
- Keep `kernel_warmup(self)` enabled.

v0.28 invokes `kernel_warmup()` from `QaicWorkerPyt.compile_or_warm_up_model()`. QAIC eager does not support the V2 model runner, so explicitly selecting V1 prevents the missing `use_v2_model_runner` state that previously caused EngineCore startup to fail.

## Dependency

Depends on #98. This PR is stacked on the v0.28 rebase and must merge after #98.

### DIFF
--- a/vllm_qaic/worker/worker.py
+++ b/vllm_qaic/worker/worker.py
@@ -202,6 +202,7 @@ class QaicWorkerPyt(QaicWorker):
             is_driver_worker=is_driver_worker,
         )
         assert self.model_config.enforce_eager
+        self.use_v2_model_runner = False
         self.parallel_config.disable_custom_all_reduce = True
         self.profiler_config = vllm_config.profiler_config
 


quic-agokhale:
Please rebase your commit to the tip of upstream dev/v0.28.0
Also run pre-commit and add sign-off

quic-agokhale:
As discussed, instead of completely skipping kernel_warmup() , set the use_v2_model_runner flag to False for QaicWorkers for now, so that it does not error out.

quic-agokhale (approved):
looks good to me!

PR #107, opened 2026-09-02
build: make RDMA support optional when building UCX

quic-vargupt:
Adding script for building ucx from source with & without RDMA/IB support


quic-vargupt:
Ready for review

ankitqti (approved):

PR #113, opened 2026-09-02
Add CLI argument --rust-image and RUST_IMAGE Dockerfile arg to allow passing user-preferred URL for pulling the Rust image

quic-agokhale:


sgoel-qc (approved):
Approved based on Apporva's feedback.

PR #114, opened 2026-09-02
[v0.23.0] Fix pending exec-object tracking for async prefill scheduling

animeshk08:
Track prefill exec-object availability via the session's queue directly instead of a separately maintained counter, which could drift from the queue's actual state. Also fix synchronize_input_prep() to only skip draining the previous batch's exec object for kv producers when a prefill exec object is still free, and always set _pending_output for async kv producers so it gets drained/completed correctly instead of leaking.


quic-gthiruko:
Tried the fix and it works as expected.

quic-sanising (approved):
LGTM

vllm_qaic/model_loader/qaic.py
```diff
@@ -673,7 +672,7 @@ def _run_pipeline_prefill(
                     if logits is not None:
                         chunk_inputs["logits"] = logits[index : index + 1]
 
-                if pending_exec_count == self.session.prefill_num_execObj:
+                if self.session.prefill_available_exec_objs.empty():
```
quic-sanising, line 675: Should this condition be not empty?

animeshk08: The condition looks ok to me, if no more prefill exec obis are available, call complete inf to free up in use ones.

PR #120, opened 2026-09-16
Adding code changes for handling rank 4 conv & recurrent states for model with GDN during KV$ transfer

quic-vargupt:


quic-vargupt:
Ready for review.

PR #121, opened 2026-09-17
Adding support for heterogenous Prefill-Decode disaggregation with Nixl

quic-vargupt:
This PR is adding the basic support for having Nvidia gpu's run prefill and we can run decode on Qaic AI100 accelerator using nixl. The KV$ from Nvidia's GPU are first staged to system DRAM and then nixl on decode pulls it using RDMA.


quic-vargupt:
This PR builds on top of PR101 and PR108

